PR #24, opened 2023-07-10
Fix font resolution in local and dist

davidicus:
**Summary**

- Fix the build config settings to account for font import resolution locally and in build folder

- added main filed to `package.json` to enable` npm link` to work properly
- removed unnecessary `rimfaf` declaration in `package.json` to account for config change
- Removed unnecessary asset from public folder (vite.svg)
- update imports in all scss files to account for new folder structure
- changed to imports for compatibility with web project until it is upgraded to use sass package
- added resolution fields to `vite.config.js` and `tsconfig.json`


adietrich3074 (dismissed):
Looks good. Just have that one question about the vars import

adietrich3074 (approved):
Looks like the vars import is needed in multiple places so ignore my previous comment

PR #50, opened 2024-04-08
L3-837 max width for viewingslist

Sisawat:
Closes #L3-837

**Summary**

Sets min and max width for ViewingsList Widget so the widget is visible on smaller screens. 

- `seldon\src\components\ViewingsList\_viewingsList.scss`

- [ ] Changes or new components should either write new or update existing documentation.

### DIFF
--- a/src/components/ViewingsList/_viewingsList.scss
+++ b/src/components/ViewingsList/_viewingsList.scss
@@ -9,7 +9,8 @@
   flex-direction: column;
   gap: 2.5rem;
   padding: 1rem;
-  width: 38rem;
+  min-width: 27rem;
+  max-width: 38rem;
 
   &__title {
     @include headerText;
--- a/src/components/ViewingsList/_viewingsList.scss
+++ b/src/components/ViewingsList/_viewingsList.scss
@@ -9,7 +9,6 @@
   flex-direction: column;
   gap: 2.5rem;
   padding: 1rem;
-  min-width: 27rem;
   max-width: 38rem;
 
   &__title {
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -17,7 +17,7 @@
 # These users own any files in the following directory at the root of
 # the repository and any of its subdirectories.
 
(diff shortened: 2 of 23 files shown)


src/components/ViewingsList/_viewingsList.scss
```diff
@@ -9,7 +9,8 @@
   flex-direction: column;
   gap: 2.5rem;
   padding: 1rem;
-  width: 38rem;
+  min-width: 27rem;
```
davidicus, line 12: Curious as to why we are setting a min size of 432px, which may not be small enough for mobile. Not that our interface is ready for that now, but at some point it would be nice to support a mobile design.

adietrich3074 (dismissed):

src/components/ViewingsList/_viewingsList.scss
Sisawat: The labels go awry if it is shrunk any smaller, and mobile was not a requirement at this time.

adietrich3074: I think it would also be a good idea to decide on a minimum screen size we support. At my last company we supported down to 320px because that was the width of the iPhone5. Maybe we could work together with the design team to look at how far back in devices we'd like to support and make it know to the business what our minimum screen size is.

davidicus: The [design system](https://www.figma.com/proto/9jVSJUHAISjJ3A3dOUuMLi/SG%3A-Grid-%26-Spacing?page-id=5793%3A989&type=design&node-id=5793-1924&viewport=851%2C1151%2C0.35&t=3nxlMYGOAYx628tx-1&scaling=contain&starting-point-node-id=5793%3A1924) supports a breakpoint of 360px - 960px.

Sisawat: Shall I remove the `min-width` or proceed to merge?

davidicus: I never sent the message. Sorry, let's remove the min width for now.

PR #91, opened 2024-05-01
L3-1468 create link component

scottdickerson:
**Jira ticket**

[L3-1468](https://phillipsauctions.atlassian.net/browse/L3-1468)

**Screenshots**
image

**Summary**

New Link component created to be used in the About Us page for the secondary navigation.  This is just the Link itself, the List of links will be an additional component.

- change(storybook): hide inherited props from HTMLElements, and switch to `react-docgen-typescript` to fix the broken source maps
- add(Link): stories and the link component
- change(utils): add `underline` mixin since the default `text-decoration: underline` didn't match the figma spec
- change(typography): fix the font face name and mixin to be specific to the woff
- change(vite): update aliases to allow components to import the type.scss file

- Test the variants of the link against the Figma component documentation and confirm the styles match
- Check the variants at different breakpoints to confirm the text adjusts everywhere except for `inline` which inherits from the main document
- Check hover and focus states for the Link
- Check the stories for External links and Custom link rendering
- Verify that source code for `Link.tsx` can be viewed in the Chrome inspector

**Regression Test**

- N/A

**Evidence of testing**

image
image

- [ ] Changes or new components should either write new or update existing documentation.

[L3-1468]: https://phillipsauctions.atlassian.net/browse/L3-1468?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Link/Link.tsx
```diff
@@ -0,0 +1,69 @@
...
+  children: React.ReactNode;
+  /**
+   * URL to navigate to when clicked
+   */
+  href: string;
+  /**
+   * Can be used to render alternative link components like the prefetching `Link` from `@remix-run/react`.
+   * If you implement this you should handle the `children`, `dataTestId`, `id`, `className`, and `href` props.
+   */
+  component?: React.ElementType<LinkProps & { 'data-testid': string }>;
```
davidicus, line 38: This was going to be `as` or `element`.  I am also not getting any TS errors for `data-testid` when omitting this in the one I did for Grid
```js
element?: React.ElementType<GridProps>;
```

src/components/Link/Link.tsx
```diff
@@ -0,0 +1,69 @@
...
+ * A component that can be used to navigate to different pages or external websites. Renders a standard anchor tag by default.
+ *
+ * [Figma Link](https://www.figma.com/file/xMuOXOAKVt5HC7hgYjF3ot/Components-v2.0?node-id=5736%3A13364&mode=dev)
+ */
+const Link = ({ children, id, className, component: Component, variant, href, ...props }: LinkProps) => {
+  const classNames = classnames(`${px}-link`, getLinkVariantClassName(variant ?? LinkVariants.standalone), className);
+  const dataTestId = id ? `link-${id}` : `link`;
+  const isExternal = href.match(/https?/);
+
+  const RootComponent = Component ?? 'a';
```
davidicus, line 53: Can we add to the default props as part of the argument declaration?

```js
const Link = ({
  children,
  id,
  className,
  element: Element = 'section',
  variant,
  href,
  ...props }: LinkProps) => {
```

src/scss/_type.scss
```diff
@@ -31,15 +36,15 @@
     font-size: $cta-sm-label-size;
     font-weight: 600;
     letter-spacing: 0.0625rem;
-    line-height: 1.25;
+    line-height: calc(1.25 * $cta-sm-label-size);
```
davidicus, line 39: Not sure why we need calc here. Without a unit isn't `1.25` the same as this?

src/scss/_type.scss
```diff
@@ -31,15 +36,15 @@
...
     text-transform: uppercase;
   }
 
   @if $label == 'cta-lg' {
     font-size: $cta-lg-label-size;
     font-weight: 700;
     letter-spacing: 0.125rem;
-    line-height: 1.25;
+    line-height: calc(1.25 * $cta-lg-label-size);
```
davidicus, line 47: Same comment as above

src/scss/_typography.scss
```diff
@@ -37,15 +37,14 @@ body {
 }
 
 @font-face {
-  font-family: DistinctDisplay;
+  font-family: DistinctDisplayItalic;
```
davidicus, line 40: Why are we changing these back? 

By doing this you would have to change the font family of any item you would want to add italics or bold? Since in Phillips-public and eventually in Remix we will have a the typography file declare all section headers to have this font we can just add styling like bold and italic through traditional means.

davidicus (changes requested):
Some minor comments

src/components/Link/Link.tsx
```diff
@@ -0,0 +1,67 @@
...
+}
+
+/**
+ * ## Overview
+ *
+ * A component that can be used to navigate to different pages or external websites. Renders a standard anchor tag by default.
+ *
+ * [Figma Link](https://www.figma.com/file/xMuOXOAKVt5HC7hgYjF3ot/Components-v2.0?node-id=5736%3A13364&mode=dev)
+ */
+const Link = ({ children, id, className, element: Component = 'a', variant, href, ...props }: LinkProps) => {
```
davidicus, line 48: I am not opposed to calling this `Component` necessarily but we already have `Element` in use and I think it is easier to make the connection between the prop and what is being used by keeping it the same name as the prop. Thoughts?

src/components/Link/Link.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const classNames = classnames(`${px}-link`, getLinkVariantClassName(variant ?? LinkVariants.standalone), className);
```
davidicus, line 49: I think we should define argument defaults. e.g. `variant = LinkVariants.standalone`. This is more explicit, I suppose. Thoughts?

davidicus (commented):
Few minor comments to discuss before merging as I would like to have a consensus on what our coding style should be.

constantinehuzenko (approved):

davidicus (dismissed):

constantinehuzenko (approved):

davidicus (approved):

PR #136, opened 2024-05-22
feat(subscribe): added validation and error to subscribe

constantinehuzenko:
**Jira ticket**

[Jira Ticket ID](https://phillipsauctions.atlassian.net/browse/L3-1664)

**Screenshots**
image
image
image

**Summary**

Adds loading and invalid states to Subscribe component, as well as native validation

- change(Subscribe.tsx): added states and way of submitting

- Subscribe component has to show `Loading...` with `subscriptionState=loading`
- Subscribe component has to show `Error` with `subscriptionState=invalid`
- Input has to show native validation when email is wrong

**Regression Test**

**Evidence of testing**

- Post logs, screenshots, etc


src/components/Subscribe/Subscribe.tsx
```diff
@@ -33,6 +33,22 @@ export interface SubscribeProps extends React.HTMLAttributes<HTMLFormElement> {
    * Subscribe title text
    */
   title?: string;
+  /**
+   * Subscribe loading text
+   */
+  loadingText?: string;
```
scottdickerson, line 39: maybe should make this a ReactNode just in case...

src/components/Subscribe/Subscribe.tsx
```diff
@@ -33,6 +33,22 @@ export interface SubscribeProps extends React.HTMLAttributes<HTMLFormElement> {
...
+  /**
+   * Subscribe error text
+   */
+  invalidText?: string;
+  /**
+   * Subscribe state for loading or error
+   */
+  subscriptionState?: 'loading' | 'invalid' | null;
```
scottdickerson, line 47: could be a good candidate for an Enum

davidicus (commented):
These changes look good. I am wondering if we should absolute position the loading text so that we do not have a shift when the loading state comes. Also, don't we need a success message as well?

src/components/Subscribe/Subscribe.tsx
```diff
@@ -63,11 +86,18 @@ const Subscribe = ({
     >
       <h3 className={`${px}-subscribe__title`}>{title}</h3>
       {blurb ? <p className={`${px}-subscribe__blurb`}>{blurb}</p> : null}
+      {isLoading ? <p className={`${px}-subscribe__loading`}>{loadingText}</p> : null}
+
       <Input
         className={`${px}-subscribe__input`}
         type="email"
```
scottdickerson, line 93: TBH, I really don't like the look of the OOTB browser validation and would rather just do our own invalid state validation

src/components/Subscribe/Subscribe.tsx
```diff
@@ -33,6 +33,22 @@ export interface SubscribeProps extends React.HTMLAttributes<HTMLFormElement> {
...
+   */
+  invalidText?: string;
+  /**
+   * Subscribe state for loading or error
+   */
+  subscriptionState?: 'loading' | 'invalid' | null;
+  /**
+   * Subscribe input change handler
+   */
+  onInputChange?: (e: React.ChangeEvent<HTMLInputElement>) => void;
```
scottdickerson, line 51: rather than track every change event to the `<input`, I think it would be better to use an `onSubscribe` action that only triggers when the form submission is successful, and passes the current state of the input

src/components/Subscribe/Subscribe.tsx
```diff
@@ -52,8 +68,15 @@ const Subscribe = ({
...
+  invalidText = '',
+  subscriptionState = null,
+  onInputChange = () => ({}),
   ...props
 }: SubscribeProps) => {
+  const isInvalid = subscriptionState === 'invalid';
+  const isLoading = subscriptionState === 'loading';
+
   return (
     <Element
```
scottdickerson, line 81: I would add the `onSubmit` handler to the parent form instead of the button's `onClick` using the form submission behavior

src/components/Subscribe/Subscribe.tsx
constantinehuzenko: Don't think so cause if we have to add spinner or something spinner won't be on that place, so we'll have to change our component anyway

src/components/Subscribe/Subscribe.tsx
constantinehuzenko: Do we have file for common enum/types?

src/components/Subscribe/Subscribe.tsx
constantinehuzenko: It's what I did, just left onClick for button just in case

src/components/Subscribe/Subscribe.tsx
scottdickerson: you can see the desired error states here: https://www.figma.com/design/xMuOXOAKVt5HC7hgYjF3ot/Components-v2.0?node-id=5730-11728&t=Nl1MOE1dSPwIxfrS-4

src/components/Subscribe/Subscribe.tsx
```diff
@@ -63,11 +102,19 @@ const Subscribe = ({
...
         type="email"
+        name="email"
         placeholder={inputPlaceholder}
         labelText={inputLabelText}
+        invalid={isInvalid}
+        invalidText={invalidText}
+        warn={warn}
+        warnText={warnText}
+        required
+        onChange={onInputChange}
```
scottdickerson, line 117: I'm guessing you're wanting to allow consumers to flag the field as invalid as the user is typing, but I don't believe that's what UX is desiring.  Instead they seem to request the error state to only be triggered on Form submission, so I'd guess this would be done in the onSubmit handler of the top-level form element.

src/components/Subscribe/Subscribe.tsx
```diff
@@ -63,11 +102,19 @@ const Subscribe = ({
...
+
       <Input
         className={`${px}-subscribe__input`}
         type="email"
+        name="email"
         placeholder={inputPlaceholder}
         labelText={inputLabelText}
+        invalid={isInvalid}
+        invalidText={invalidText}
+        warn={warn}
```
scottdickerson, line 114: I don't see any warn UX in the spec for the Input component here: https://www.figma.com/design/xMuOXOAKVt5HC7hgYjF3ot/Components-v2.0?node-id=5730-11728&t=Nl1MOE1dSPwIxfrS-4

scottdickerson (approved):
LGTM

src/components/Input/_input.scss
davidicus, line 1: Why are these changes added. These inputs were built to design spec. Also, if you look at the input story and put it in a warn state you will see the text overlays the input.

davidicus (changes requested):

src/components/Input/_input.scss
constantinehuzenko: I added this to prevent the shift we experienced when the input changes from the default state to the loading or invalid state (https://github.com/PhillipsAuctionHouse/seldon/pull/136#pullrequestreview-2071955632). You're right that with two or more lines, we have an overlay on the input. The problem is, to avoid that overlap, we'd either need to add more space between the input and the button, or remove the absolute positioning completely, which would bring back the shift issue.

davidicus: Hmm. I wonder if we add a bottom margin that is removed when in an invalid or warning state?

constantinehuzenko: I removed that for now so I can move forward and finish footer, after that we can think how we proceed with input, could you resolve this one since I removed css which was causing problems?

davidicus (approved):

PR #148, opened 2024-05-28
L3 2053 Add new text component

scottdickerson:
**Jira ticket**

[L3-2053](https://phillipsauctions.atlassian.net/browse/L3-2053)

**Screenshots**
image

**Summary**

Adds new Text component to create semantic HTML elements and apply the correct text tokens

- change(storybook): upgrade the storybook to 8.1.4 and unify the autodocs tag into one location
- change(DatePicker): fix missing async checks causing intermittent test failures
- change(HeroBanner): Add margin bottom
- change(type.scss/text-showcase): change the 'text' to 'label' for label tokens
- add(Text): Add new text component that based on the variant applies the correct styles and semantic HTML

- Flip through the different variants of the Text story and check their margins, HTML rendering and classes against the Text Token Showcase story

**Regression Test**

- Verify the Text Token showcase still renders correctly (specially for labels).
- Validate that the existing stories still show the documentation that autodocs was rendering

**Evidence of testing**
Autodocs
image

image

New Components

[L3-2053]: https://phillipsauctions.atlassian.net/browse/L3-2053?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


scottdickerson:
@RyanOlanoPHILLIPS I've changed the token from `$text` to `$label` to be more explicit

scottdickerson:
@davidicus do you think I should put the bottom margins into the tokens themselves instead of just in the Text component?

package.json
```diff
@@ -42,21 +42,22 @@
     "flatpickr": "^4.6.13"
   },
   "devDependencies": {
-    "@chromatic-com/storybook": "^1.4.0",
+    "@chromatic-com/storybook": "^1.5.0",
     "@commitlint/cli": "^19.3.0",
     "@commitlint/config-conventional": "^19.2.2",
+    "@esbuild/linux-x64": "^0.21.4",
```
davidicus, line 48: What is this package for?

src/components/Text/Text.tsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
+import { px } from '../../utils';
+import classNames from 'classnames';
```
davidicus, line 3: Can we standardize the import for this? I think we have more `classnames` for this package import, which I prefer because it matches the package and does not get confused with the attribute.

davidicus (commented):
so far so good. Have to see why the build is failing so we can look at the stories. Just a small comment about the classname import. 

As far as adding the margin to the token, that may make sense. But if we are using the text component everywhere maybe leaving the tokens without offers an escape hatch?

davidicus (approved):

PR #154, opened 2024-05-31
fix(build): L3-2289 - add entry for utils

davidicus:
**Jira ticket**

[L3-2289](https://phillipsauctions.atlassian.net/browse/L3-2289)

**Screenshots**
| Before | After |
| ------ | ----- |
| image| image |

**Summary**

Import the the utils file into our index entry file.

- Change import to include other util functions. 

- Run the build command locally and make sure you see the `emailValidation` function being exported. 

New Components

[L3-2289]: https://phillipsauctions.atlassian.net/browse/L3-2289?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -26,7 +26,7 @@ export {
 export * from './components/Text';
 
 // Utils
-export { PaddingTokens } from './utils';
+export * from './utils';
 
 // 📑 Pages
 export { default as Page } from './pages/Page';
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -27,7 +27,7 @@ export default defineConfig({
       name: 'seldon',
     },
     rollupOptions: {
-      input: 'src/index.ts',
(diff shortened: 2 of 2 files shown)


scottdickerson (commented):
why did you decide to do it this way instead of just adding the utils export to `index.ts`?

davidicus:
> why did you decide to do it this way instead of just adding the utils export to `index.ts`? I think going forward we might want to have multiple other subdirectories exported and I'm not sure that we would want to change the vite.config.ts for each one..

I am not super opinionated on this. I guess, I would rather config not be updated often. Good call, I will change.

davidicus:
I used a star import here. Maybe anything in this utils folder should be a utility that can be broadly applied. And then any scoped utils are done like we did for link.

PR #167, opened 2024-06-06
feat(modal): L3-2144 added modal and updated button

constantinehuzenko:
**Jira ticket**

[L3-2144](https://phillipsauctions.atlassian.net/browse/L3-2144)

**Screenshots**
image

**Summary**

This pull request introduces a new `Modal` component in the file `Modal.tsx`. The Modal component is a reusable UI element that can be used to display content in a layer above the main application.

- Modal: added component, story and styles
- Button: changed styles

**Regression Test**

**Evidence of testing**

- Post logs, screenshots, etc

New Components

[L3-2144]: https://phillipsauctions.atlassian.net/browse/L3-2144?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,55 @@
+import classnames from 'classnames';
+import { px } from '../../utils';
+import CloseIcon from '../../assets/close.svg?react';
+import Button from '../Button/Button';
+
+export interface ModalProps extends React.HTMLAttributes<HTMLElement> {
+  /**
+   * Boolean that will determine if the modal is open
+   */
+  open?: boolean;
```
scottdickerson, line 10: I think that any boolean property should follow a naming convention like `isOpen` https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/boolean-prop-naming.md.  Can you turn this eslint rule on or open an Jira ticket to track that?

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,55 @@
...
+  /**
+   * Function to close the modal
+   */
+  onClose?: () => void;
+}
+
+const onClosePlaceholder = () => {
+  console.log('On close clicked');
```
scottdickerson, line 18: I'd rather this be a noop instead of a local function.  I think we have a function like this defined in a shared utils file already @davidicus

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,55 @@
...
+    return null;
+  }
+
+  const buttonClick = (e: React.MouseEvent) => {
+    e.stopPropagation();
+    onClose();
+  };
+
+  return (
+    <div className={classnames(`${px}-modal`)} onClick={onClose} data-testid="modal-background">
```
scottdickerson, line 39: I don't think this passes a11y as it's not setting aria stuff correctly.  Have you considered using this library https://www.radix-ui.com/primitives/docs/components/dialog to get the accessibility right?  Or if that seems like overkill this one which is lighter weight? https://reactcommunity.org/react-modal/

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,55 @@
...
+      <div className={classnames(`${px}-modal__container`)} data-testid="modal-container" {...props}>
```
scottdickerson, line 40: we need to extract the classname prop and add it to the top-level className so that if they pass it, it doesn't override our `{$px}-modal` className

```
<div className={classnames(`${px}-modal`, className)} onClick={onClose} data-testid="modal-background" {...props}>
```

In addition, I think we should always spread the props to the top-level component like above, not a child component

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,55 @@
+import classnames from 'classnames';
+import { px } from '../../utils';
+import CloseIcon from '../../assets/close.svg?react';
+import Button from '../Button/Button';
+
+export interface ModalProps extends React.HTMLAttributes<HTMLElement> {
```
scottdickerson, line 6: you can make this more specific:


```suggestion
export interface ModalProps extends React.HTMLAttributes<HTMLDivElement> {
```

scottdickerson: you might also choose to override the children prop to make it explicit that it's required.

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,55 @@
...
+    onClose();
+  };
+
+  return (
+    <div className={classnames(`${px}-modal`)} onClick={onClose} data-testid="modal-background">
+      <div className={classnames(`${px}-modal__container`)} data-testid="modal-container" {...props}>
+        <Button
+          className={classnames(`${px}-modal__close`)}
+          data-testid="modal-button"
+          buttonType="icon"
```
scottdickerson, line 44: Nice!

src/components/Button/Button.tsx
```diff
@@ -18,7 +18,7 @@ export interface ButtonProps extends React.HTMLAttributes<HTMLButtonElement> {
   /**
    * Is this the principal call to action on the page?
    */
-  buttonType?: 'primary' | 'secondary' | 'ghost';
+  buttonType?: 'primary' | 'secondary' | 'ghost' | 'icon';
```
scottdickerson, line 21: it would be nice if we renamed this to `variant` while we were in here... :|

scottdickerson: this is a little bit confusing with the `iconLast` options above...  I guess, do we need that prop still?  Maybe we can ask David what he thinks...  I also think it's a bit confusing to have the story be button with icon.

scottdickerson (approved):
this needs to be updated in the Button.stories to allow selecting this new variant. TBH this should really be a variant enum and use the same approach we use in the `Link` story so it always updates.

```
    options: ['primary', 'secondary', 'ghost'],
```

scottdickerson (approved):
I think you forgot to export the new Modal.tsx component and Props from the main `index.ts`

src/components/Button/_button.scss
```diff
@@ -60,6 +60,21 @@
...
+    background-color: unset;
+    border: unset;
+    min-width: unset;
+
+    & > svg {
+      margin-inline-end: unset;
+      width: unset;
+    }
+
+    &:hover {
```
scottdickerson, line 73: hmm, I actually think the icon maybe should have a hover state and definitely a pointer to indicate that it's actionable.  The icon button doesn't really seem to be working for me in our stories

scottdickerson (changes requested):
I think some changes below should be done before merging

package.json
```diff
@@ -39,7 +39,8 @@
   },
   "dependencies": {
     "classnames": "^2.5.1",
-    "flatpickr": "^4.6.13"
+    "flatpickr": "^4.6.13",
+    "react-modal": "^3.16.1"
```
davidicus, line 43: @scottdickerson @constantinehuzenko I am fine with using react-modal since it is pretty widely used but since we were looking at [this](https://www.radix-ui.com/primitives) library for primitives do we want to use this library to build our modal instead?

src/components/Button/Button.tsx
davidicus: I like renaming to variant. Open to suggestions on icon placement prop.

src/components/Modal/Modal.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  it('calls the onClose function when the close button is clicked', () => {
+    render(
+      <Modal isOpen={true} onClose={onCloseMock}>
+        <div>Modal Content</div>
+      </Modal>,
+    );
+
+    const closeButton = screen.getByLabelText('Close Modal');
+    fireEvent.click(closeButton);
```
davidicus, line 43: minor nitpick since it has no difference here but should we standardize on using `user-event` over `fireEvent`?

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,54 @@
...
+import CloseIcon from '../../assets/close.svg?react';
+import Button from '../Button/Button';
+import ReactModal from 'react-modal';
+
+export interface ModalProps extends ReactModal.Props {
+  /**
+   * Function to close the modal
+   */
+  onClose?: () => void;
+}
```
davidicus, line 12: Do we want to describe the `isOpen` prop in our interface here for documentation?

davidicus (commented):
Looking great. Just a few comments I would like to discuss.

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,58 @@
...
+ *
+ */
+
+const Modal = ({ children, className, isOpen = false, onClose = noOp, ...props }: ModalProps) => {
+  if (!isOpen) {
+    return null;
+  }
+
+  const buttonClick = (e: React.MouseEvent) => {
+    e.stopPropagation();
```
scottdickerson, line 31: why do we need to stopPropagation here?  Parent components just might wanna listen

scottdickerson (changes requested):
When I switch the Button -> Playground story to `icon` the content of the button disappears:
image

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,54 @@
...
+   * Boolean to determine if the modal is open
+   */
+  isOpen: boolean;
+  /**
+   * Function to close the modal
+   */
+  onClose?: () => void;
+}
+
+ReactModal.setAppElement('body');
```
scottdickerson, line 18: I think this won't work in an accessible way:
https://github.com/reactjs/react-modal/issues/576#issuecomment-451789536

I think we might need to require that the `appElement` prop be passed to modal by consumers so that they can indicate which element to hide.  Ideally we'd have a common root element approach like `#root` in both Remix and Phillips-public, but I don't see either

scottdickerson (approved):
LGTM

PR #190, opened 2024-06-20
fix(booleanProps): L3-2356 eslint rules to enforce boolean props

scottdickerson:
**Jira ticket**

[L3-2356](https://phillipsauctions.atlassian.net/browse/L3-2356)

**Summary**

We need to enhance our eslint rules to catch more errors at coding time.  Specifically I want to add some more best practice React ones, async/await promise enforcement and validation of boolean property naming conventions

- change(eslint): add react/recommended rules, `react/boolean-prop-naming`, `require-await` and `@typescript-eslint/no-floating-promises`
- change(Button*/ViewingsListCard): I did need to rename a few props but I have noticed that no external consumers are using them yet in `phillips-public` or `phillips-public-remix` so I think we can get away with that small change without putting out the major version.

- Validate if you try and write code that would fail any of those rules ,it fails linting (check the web for some good examples of how to violate those rules)

**Regression Test**

- N/A

**Evidence of testing**

image

image

New Components

[L3-2356]: https://phillipsauctions.atlassian.net/browse/L3-2356?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,2 +1,3 @@
 /dist
 /storybook-static
+vite.config.ts
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -6,17 +6,32 @@ module.exports = {
   },
   extends: [
     'eslint:recommended',
+    'plugin:react/recommended',
+    'plugin:react/jsx-runtime',
     'plugin:@typescript-eslint/recommended',
     'plugin:react-hooks/recommended',
     'plugin:storybook/recommended',
   ],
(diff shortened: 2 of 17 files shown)


davidicus:
Documenting some squirliness with some variations of the trailing promises, but I have verified that each of these rules work. 

image

davidicus (approved):

PR #196, opened 2024-06-26
feat(components): L3-2559 Build Accordion component in Seldon library

zhenpanda:
**Jira ticket**

(https://phillipsauctions.atlassian.net/browse/L3-2559)

**Screenshots**

https://github.com/PhillipsAuctionHouse/seldon/assets/2293731/a9baf4fc-f834-4798-8f24-d328539409d6

**Summary**

Create an Accordion component in Seldon to spec from [this Figma](https://www.figma.com/design/xMuOXOAKVt5HC7hgYjF3ot/Components-v2.0?node-id=4433-163013&m=dev).

**Regression Test**

**Evidence of testing**

- Post logs, screenshots, etc

New Components


Sisawat (approved):

src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import classnames from 'classnames';
+import { px } from '../../utils';
```
davidicus, line 3: I believe your baseClassname will have the `px` prefix and you will not have to write your own here. If you look in the inspector you will see the `phillips-accordion` being applied twice.

src/components/Accordion/AccordionItem.tsx
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * ## Overview
+ *
+ * A single Accordion Item in a list
+ */
+
+const AccordionItem = ({ ...props }: AccordionItemType) => {
+  getCommonProps(props, 'AccordionItem');
+  const { locked, lockedVariation, lockedContent, variation, label, content, isLastItem, id } = props;
+  const [contentState, setContentState] = useState('collapsed');
```
davidicus, line 20: Would be better to be a boolean?

```suggestion
  const [isCollapsed, setIsCollapsed] = useState('false');
```

src/components/Accordion/AccordionItem.tsx
```diff
@@ -0,0 +1,69 @@
...
+        ? 'lockBlueIcon'
+        : contentState === 'collapsed'
+          ? 'plusIcon'
+          : 'minusIcon');
+
+  return (
+    <div className={classNames(`${px}-accordionItem`, !isLastItem && `${px}-accordionItem__border_bottom`)}>
+      <div
+        className={classNames(`${px}-accordionItem__label`, !locked && `${px}-accordionItem__label_hoverable`)}
+        onClick={() => !locked && setContentState(contentState === 'collapsed' ? 'expanded' : 'collapsed')}
```
davidicus, line 36:
```suggestion
        onClick={() => !locked && setIsCollapsed(prevState => !prevState)}
```

src/components/Accordion/AccordionItem.tsx
```diff
@@ -0,0 +1,69 @@
...
+      >
+        <div
+          className={classNames(`${px}-accordionItem__label_text`, isLargeVariation && `${px}-accordionItem__label_lg`)}
+        >
+          {label}
+        </div>
+        <img
+          className={classNames(`${px}-accordionItem__label_icon`, isLargeVariation && `${px}-accordionItem__icon_lg`)}
+          src={
+            locked ? (lockedVariation ? lockBlueIcon : lockIcon) : contentState === 'collapsed' ? plusIcon : minusIcon
```
davidicus, line 47:
```suggestion
            locked ? (lockedVariation ? lockBlueIcon : lockIcon) : isCollapsed ? plusIcon : minusIcon
```

src/components/Accordion/AccordionItem.tsx
```diff
@@ -0,0 +1,69 @@
...
+          data-testid={iconId}
+        />
+      </div>
+
+      {locked && lockedContent ? (
+        <div>{lockedContent}</div>
+      ) : (
+        <div
+          className={classNames(
+            contentState === 'collapsed' ? `${px}-accordionItem__collapsed` : `${px}-accordionItem__expanded`,
```
davidicus, line 58: We probably do not need to have both class states? If it can only be collapsed or expanded I would probably just have one be the default style and then when the other class is applied we overwrite those styles.

src/components/Accordion/AccordionItem.tsx
```diff
@@ -0,0 +1,69 @@
+import { px } from '../../utils';
+import { getCommonProps } from '../../utils';
+import { AccordionItemType } from './types';
+import plusIcon from '../../assets/plus.svg';
+import minusIcon from '../../assets/minus.svg';
+import lockIcon from '../../assets/lock.svg';
+import lockBlueIcon from '../../assets/lockBlue.svg';
+import { useState } from 'react';
+import classNames from 'classnames';
```
davidicus, line 9: import should be `classnames` to avoid confusion with the prop/attribute

src/components/Accordion/AccordionItem.tsx
```diff
@@ -0,0 +1,69 @@
...
+        />
+      </div>
+
+      {locked && lockedContent ? (
+        <div>{lockedContent}</div>
+      ) : (
+        <div
+          className={classNames(
+            contentState === 'collapsed' ? `${px}-accordionItem__collapsed` : `${px}-accordionItem__expanded`,
+            isLargeVariation && `${px}-accordionItem__child_text_lg`,
```
davidicus, line 59: this should be written as the following
```suggestion
           {[`${px}-accordionItem__child_text_lg`]:  isLargeVariation}
```

src/components/Accordion/_accordion.scss
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  &__collapsed {
+    display: none;
+    height: auto;
+  }
+
+  &__child_text_lg {
+    font-size: 20px;
+    margin: 20px 0;
```
davidicus, line 70: should we be using tokens here? Also would like to use `rem` over `px` whenever possible

src/components/Accordion/types.ts
```diff
@@ -0,0 +1,16 @@
+// Add any additional typescript types here
+interface AccordionItemType {
```
davidicus, line 2: This should be defined in the the AccordionItem file

src/components/Accordion/utils.test.ts
davidicus, line 1: We can delete if no utils were needed for this component

src/components/Accordion/utils.ts
davidicus, line 1: Same comment: delete if not used

src/index.ts
davidicus, line 1: I think we need to add `AccordionItem` here

davidicus (changes requested):
Great start and awesome turn around. We should be starting our components out by looking at the Radix library to see if there are any primitives that we can use for a11y reasons as mentioned in our [contributing guidelines]( [isLargeVariation](https://github.com/PhillipsAuctionHouse/seldon/blob/main/src/docs/CONTRIBUTING.md#starting-a-new-component)). Otherwise we will have a lot of a11y stuff that we will have to add here.

src/components/Accordion/AccordionItem.tsx
```diff
@@ -0,0 +1,83 @@
...
+  isLastItem?: boolean;
+  isLocked: boolean;
+  isLockedVariation?: boolean;
+  // eslint-disable-next-line  @typescript-eslint/no-explicit-any
+  lockedContent?: any;
+  variation: 'lg' | 'sm';
+  label: string;
+  // eslint-disable-next-line  @typescript-eslint/no-explicit-any
+  content?: any;
+  hasTransition?: boolean;
```
davidicus, line 21: When writing out the interface you should have comments above each one. This is what gets populated in our docs for storybook. You can look at the [grid component](https://github.com/PhillipsAuctionHouse/seldon/blob/ef7ee39ff5f7d8fe9286ba4b0bcb0cd9bcbe9556/src/components/Grid/Grid.tsx#L5) for example.

I do believe it needs to be a multiline comment e.g.

```
/**
* Your comment becomes documentation for the prop
*/

File: src/components/Accordion/AccordionItem.tsx [hunk 11]
```diff
@@ -0,0 +1,83 @@
...
+          data-testid={iconId}
+        />
+      </div>
+
+      {isLocked && lockedContent ? (
+        <div>{lockedContent}</div>
+      ) : (
+        <div
+          className={classnames(
+            hasTransition ? `${px}-accordionItem__transition__default` : `${px}-accordionItem__default`,
```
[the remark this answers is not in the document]
davidicus: Can the base classname be the default?

src/components/Accordion/AccordionItem.tsx
```diff
@@ -0,0 +1,69 @@
...
+import classNames from 'classnames';
+
+/**
+ * ## Overview
+ *
+ * A single Accordion Item in a list
+ */
+
+const AccordionItem = ({ ...props }: AccordionItemType) => {
+  getCommonProps(props, 'AccordionItem');
```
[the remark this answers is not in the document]
davidicus: By this I mean we should be using the getCommonProps.

src/components/Accordion/_accordion.scss
```diff
@@ -0,0 +1,92 @@
...
+    transition: 300ms grid-template-rows ease;
+  }
+
+  &__transition__default > div {
+    overflow: hidden;
+  }
+
+  &__transition__collapsed > div:first-child {
+    margin: 0 !important;
+    padding: 0 !important;
```
davidicus, line 85: We should not have to use `!important`

davidicus (changes requested):
Let's meet and review tomorrow. I think we would be better suited starting with a Radix primitive. There are a lot of a11y concerns that are not addressed here that we would hope to get from the primitive library.

src/assets/lock.svg
adietrich3074, line 1: Is there a reason this is red? The figma shows blue and dark grey locks but not red.

adietrich3074 (changes requested):
Functionally and style wise I think the component looks and works great. I am concerned though that the way this component is structured deviates a lot from how we are currently building out components. I know it may not be ideal but can we meet to discuss our standards for how a component should be structured. I think we would benefit from consistency in this aspect of our component development. @davidicus @scottdickerson @constantinehuzenko

src/assets/lock.svg
zhenpanda: it's red here 
Screenshot 2024-07-10 101816

adietrich3074: Can we verify with Ryan that this is correct? I haven't seen this color used anywhere else but here and in the other examples in the figma the lock is our standard dark grey color. If Ryan says this is correct then I'm fine with it staying this way

src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const { className: baseClassName, ...commonProps } = getCommonProps(props, 'Accordion');
+  return (
+    <Accordion.Root
+      className={classnames(`${baseClassName}`, className)}
+      type={type}
+      {...commonProps}
+      id={props?.id}
+      collapsible
+    >
+      {items?.length > 0 &&
```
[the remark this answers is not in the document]
zhenpanda: if `items?.length > 0` is removed it will break `runCommonTests`

src/assets/lock.svg
zhenpanda: Yea Ryan just confirmed it's grey. I just updated the icon.
Screenshot 2024-07-11 170427

src/assets/lockBlue.svg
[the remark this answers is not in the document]
zhenpanda: `fill` value didn't work for me but `filter` did work

davidicus:
@adietrich3074 can you re-review here?

adietrich3074 (approved):

adietrich3074:
Approved 👍

PR #205, opened 2024-07-02
feat(components): L3-2507 build breadcrumb component

Zach-OfAllTrades:
**Jira ticket**

[L3-2507](https://phillipsauctions.atlassian.net/browse/L3-2507)

**Screenshots**
Static State
image

Hover State
image

**Summary**

Created a Breadcrumb component to spec from [this Figma](https://www.figma.com/design/xMuOXOAKVt5HC7hgYjF3ot/Components-v2.0?node-id=4530-104045&m=dev).

No Radix Primitive for breadcrumb but follow ARIA breadcrumb pattern [here](https://www.w3.org/WAI/ARIA/apg/patterns/breadcrumb/examples/breadcrumb/)

Created Breadcrumb & BreadcrumbItem components, related story and tests. Added chevronNext icon. Added $keyline-gray to _vars.scss.

See screenshots & code changes. 

**Evidence of testing**

image

New Components

[L3-2507]: https://phillipsauctions.atlassian.net/browse/L3-2507?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,42 @@
...
+  return (
+    <nav aria-label="Breadcrumb" className={classnames(baseClassName, className)} {...commonProps} {...props}>
+      <ol>
+        {items &&
+          items.map((item: BreadcrumbItemProps, index: number) => (
+            <BreadcrumbItem
+              href={item.href}
+              label={item.label}
+              isCurrent={items.length - 1 === index}
+              key={`${index}-breadcrumb-item`}
```
constantinehuzenko, line 34: We shouldn't use `index` as a `key`, I would suggest adding hardcoded `key` to each item inside `items` or  just use `label` as a `key` if it's unique 

"This is because React uses keys to determine whether a component needs to be updated or not. If the keys are based on the index and the order changes, React might end up re-rendering more components than necessary, negatively impacting your app's performance." - https://www.dhiwise.com/post/react-lists-and-keys-the-key-to-efficient-rendering

src/components/Breadcrumb/BreadcrumbItem.tsx
```diff
@@ -0,0 +1,40 @@
...
+const BreadcrumbItem = ({ className, href, label, isCurrent = false, ...props }: BreadcrumbItemProps) => {
+  const { className: baseClassName, ...commonProps } = getCommonProps(props, 'Breadcrumb');
+  const extraClass = !isCurrent ? 'previous' : 'current';
+
+  return (
+    <li>
+      <a
+        aria-current={!isCurrent ? false : 'page'}
+        className={classnames(baseClassName, className, extraClass)}
+        href={!isCurrent ? href : ''}
```
constantinehuzenko, line 30: Optional: we can move expressions like `!isCurrent ? false : 'page'` and  `!isCurrent ? href : ''` to their own variables

zhenpanda (approved):
lgtm

constantinehuzenko (commented):
One change is require, others are optional, great code overall!

constantinehuzenko:
Can we check with a design team how to make it look better on mobile phone?
Maybe we can add `...` so words are not that long

image

Zach-OfAllTrades:
> Can we check with a design team how to make it look better on mobile phone? Maybe we can add `...` so words are not that long Not sure if you have separate ticket for mobile image

Good catch, I was thinking about this in general with a really long drill down but I wasn't sure it was a realistic use case. But yeah here on mobile we should probably have something that looks better. I do not personally have a mobile ticket assigned.

src/components/Breadcrumb/Breadcrumb.tsx
Zach-OfAllTrades: Thanks for the insight, will update!

Sisawat (approved):

src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,45 @@
...
+            <BreadcrumbItem
+              href={item.href}
+              label={item.label}
+              isCurrent={items.length - 1 === index}
+              key={item.label}
+            />
+          ))
+        ) : (
+          <></>
+        )}
```
Zach-OfAllTrades, line 39: @constantinehuzenko is this what you had in mind with the ternary?

davidicus: you could just initialize it with an empty array as the argument default and forget the check altogether. This would probably be cleaner.

src/components/Breadcrumb/BreadcrumbItem.tsx
```diff
@@ -0,0 +1,42 @@
...
+  isCurrent?: boolean;
+  /**
+   * Label to display in the breadcrumb
+   */
+  label?: string;
+}
+
+const BreadcrumbItem = ({ className, href, label, isCurrent = false, ...props }: BreadcrumbItemProps) => {
+  const { className: baseClassName, ...commonProps } = getCommonProps(props, 'Breadcrumb');
+  const extraClass = !isCurrent ? 'previous' : 'current';
```
davidicus, line 23: We should be following the BEM class naming convention.
```
`${baseClassName}--current`
```
We should also use classnames package to do the conditional adding of classes. 
```
className={classnames(
    baseClassName,
    className, 
   { [`${baseClassName}--current`]: isCurrent }
)}
```
And since only one can be current, do we think it make sense to have the default styles include the styles of the `previous` class and then the `current` modify can overwrite those styles?

src/components/Breadcrumb/BreadcrumbItem.tsx
```diff
@@ -0,0 +1,42 @@
...
+    <li>
+      <a
+        aria-current={ariaCurrent}
+        className={classnames(baseClassName, className, extraClass)}
+        href={currentHref}
+        {...commonProps}
+      >
+        {label}
+      </a>
+      {!isCurrent && <img className={classnames(baseClassName, className, 'chevron')} src={chevronNext} />}
```
davidicus, line 37: We do prefer the explicit ternary here over `&&` as @constantinehuzenko has mentioned.

src/components/Breadcrumb/_breadcrumb.scss
```diff
@@ -0,0 +1,35 @@
+@import '#scss/_vars';
+
+.#{$px}-breadcrumb {
+  @include text($body3);
+
+  font-weight: 400;
```
davidicus, line 6: is this weight not applied by the text mixin already?

src/components/Breadcrumb/_breadcrumb.scss
```diff
@@ -0,0 +1,35 @@
...
+  ol {
+    display: flex;
+    flex-direction: row;
+
+    li {
+      display: flex;
+      flex-direction: row;
+
+      a {
+        margin: 10px;
```
davidicus, line 17: Does the figma reference spacing tokens for this?

src/components/Breadcrumb/BreadcrumbItem.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+const BreadcrumbItem = ({ className, href, label, isCurrent = false, ...props }: BreadcrumbItemProps) => {
+  const { className: baseClassName, ...commonProps } = getCommonProps(props, 'Breadcrumb');
+  const extraClass = !isCurrent ? 'previous' : 'current';
+  const ariaCurrent = !isCurrent ? false : 'page';
+  const currentHref = !isCurrent ? href : '';
+
+  return (
+    <li>
+      <a
```
davidicus, line 29: Could we utilize our [link](https://deploy-preview-205--phillips-seldon.netlify.app/?path=/story/components-links-link--playground) component here? this may make styling the hover styles unnecessary.

src/components/Breadcrumb/_breadcrumb.scss
```diff
@@ -0,0 +1,35 @@
...
+
+  ol {
+    display: flex;
+    flex-direction: row;
+
+    li {
+      display: flex;
+      flex-direction: row;
+
+      a {
```
davidicus, line 16: These styles may not be necessary if we use the link component

src/scss/_vars.scss
```diff
@@ -13,6 +13,7 @@ $white: #fff;
 // Secondary color palette
 $soft-black: #545454;
 $dark-gray: #7b7474;
+$keyline-gray: #8f8989;
```
davidicus, line 16: your color is being overwritten by line 19. This color token already exist. If the figma is referencing another color value we should call that out in the ticket but not modify this token until verifying that all uses of keyline-gray are now changing.

davidicus (changes requested):
Just a few minor comments.

src/components/Breadcrumb/Breadcrumb.tsx
constantinehuzenko: @davidicus you mean `items = []`? actually great idea!

src/scss/_vars.scss
Zach-OfAllTrades: Oh shoot, I missed that, I did a search for the hex code, not the name. This is the same problem I had with sale lot redesign having two Soft-Black's defined in figma. Who should I reach out to to confirm?

src/components/Breadcrumb/BreadcrumbItem.tsx
Zach-OfAllTrades: I originally used the Link component but I thought it would be cleaner to use an anchor component because there were several styling overrides I was going to need to do anyway. This component's hover style differs from the link component, I only saw underlined variations. I will look into it again, though.

src/components/Breadcrumb/_breadcrumb.scss
Zach-OfAllTrades: I am not seeing any spacing tokens and do not see any property when I drill down to the component on Figma. This was me eyeballing it, should I reach out to get an agreed upon spacing?

Zach-OfAllTrades: $margin-sm looks to be the closest.

Zach-OfAllTrades: Confirmed with Ryan that $margin-sm at 1rem is correct

constantinehuzenko (approved):
LGTM

src/scss/_vars.scss
```diff
@@ -13,6 +13,7 @@ $white: #fff;
 // Secondary color palette
 $soft-black: #545454;
 $dark-gray: #7b7474;
+$soft-black-2: #535050;
```
Zach-OfAllTrades, line 16: @davidicus This is the color I confirmed with Ryan, apparently it was incorrect in Figma. This is the same issue I came across when working on Sale Lot Redesign where 2 separate "soft black"s were defined in Figma.

src/components/Breadcrumb/BreadcrumbItem.tsx
adietrich3074: I'm slightly torn here. I do think we should be using the link component here for consistency but the default link component would require style overrides. Maybe this is another case for a LinkVariant like we have in the header and some other places. What do you think @davidicus?

davidicus: Does the list variant work here?

Zach-OfAllTrades: Was just looking into this, I missed the link variation first time around. This gets it somewhat close but there would still need to be several overrides to make it consistent with Figma. The font, capitalization, and color all need to be overridden and looks like when I attempt to do so I mess up the hover styling. Maybe I am doing it incorrectly? @davidicus @adietrich3074 

```
.#{$px}-breadcrumb {  
  .#{$px}-text{
    @include text($body3);
  }
  ...
}
```

^ seems to break the hover styling.

davidicus: Let's not fight it if there are too many over rides. I do think we should speak with design about these discrepancies but we can get this in and iterate after the fact.

PR #218, opened 2024-07-11
feat(types): L3-2525 added error type to SubscriptionState

constantinehuzenko:
**Jira ticket**

[L3-2525](https://phillipsauctions.atlassian.net/browse/L3-2525)

**Screenshots**
| Before | After |
| ------ | ----- |
| image | image |

**Summary**

Added error type to `SubscriptionState`

- feat(src/components/Subscribe/Subscribe.tsx): added new prop and changed the invalid state handling
- feat(src/components/Subscribe/types.ts): added new type

- Check that when subscriptionState is error, the text 'An error occurred. Please try again.' is visible and the input is red 

**Regression Test**

**Evidence of testing**

- Post logs, screenshots, etc

New Components

[L3-2525]: https://phillipsauctions.atlassian.net/browse/L3-2525?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Subscribe/Subscribe.tsx
+++ b/src/components/Subscribe/Subscribe.tsx
@@ -43,6 +43,10 @@ export interface SubscribeProps extends React.HTMLAttributes<HTMLFormElement> {
    */
   invalidText?: string;
   /**
+   * Subscribe invalid text
+   */
+  errorText?: string;
+  /**
    * Subscribe success text
    */
   successText?: string;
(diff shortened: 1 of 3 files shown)


src/components/Subscribe/Subscribe.tsx
```diff
@@ -73,24 +77,28 @@ const Subscribe = ({
   title = 'Subscribe to Newsletter',
   loadingText = 'Loading...',
   invalidText = '',
+  errorText = 'An error occurred. Please try again.',
```
scottdickerson, line 80: should the `subscriptionState` prop be changed to an object with `{ state: SubscriptionState, message: 'An error occurred', ...}`

This would reduce the number of props and keep us from having to add a different text Prop everytime we needed to add a new state.

constantinehuzenko: great idea, but I'll have to change that in Phillips Public as well so it will require more time, can we add another ticker for that? @scottdickerson

scottdickerson (approved):
LGTM

PR #234, opened 2024-07-18
fix(header): L3-2815 fix minor header style issues

scottdickerson:
**Jira ticket**

[L3-2815](https://phillipsauctions.atlassian.net/browse/L3-2815)

**Screenshots**
| Before | After |
| ------ | ----- |
| description | Missing Bottom Border and overlaying the page content |
| image | image |
| description | Scroll bar and alignment issue on mobile menu |
| image | image |
| description | Center the view upcoming auctions |
| image | image |

**Summary**

Final polish bug fixes and style fixes for the Header 1.0 component.

- change(Header): wrap the Logo in an anchor tag with translated label
- change(*.scss): make utility function to determine desktop or mobile and a variable for user-management-height and header-height
- change(header.scss): make sticky
- change(input.scss): make cursor pointer for radio input
- change(navigationItem): use new offwhite color background. fix issue where the non-expanded menu items disappeared
- change(navigationList): only scroll if necessary, fix cursor pointer that was making too many things look clickable
- change(UserManagement): put special cursor pointer above Login span and account icon

- Confirm fixes listed above in the storybook story

**Regression Test**

- Check both the mobile header and desktop header to ensure existing functionality works well.

**Evidence of testing**

- See images above

New Components

[L3-2815]: https://phillipsauctions.atlassian.net/browse/L3-2815?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Header/_header.scss
+++ b/src/components/Header/_header.scss
@@ -9,7 +9,6 @@
   padding: $user-management-height $spacing-md 0; // save space for the user management
   position: sticky;
   top: 0;
-  width: 100vw;
 
   @include isHeaderMobile {
     padding: 0;
--- a/src/components/Header/_header.scss
+++ b/src/components/Header/_header.scss
@@ -1,17 +1,18 @@
 @import '#scss/vars';
+@import '#scss/utils';
 
 .#{$px}-header {
   align-items: center;
   background-color: $white;
   border-bottom: 1px solid rgba(0, 0, 0, 10%);
   display: flex;
-  height: 60px;
-  padding: 0 $spacing-md;
-  position: fixed;
-  top: 30px;
+  padding: $user-management-height $spacing-md 0; // save space for the user management
+  position: sticky;
+  top: 0;
   width: 100vw;
 
-  @media (max-width: #{$breakpoint-md - 1}) {
+  @include isHeaderMobile {
+    padding: 0;
     top: 0;
   }
 
@@ -22,7 +23,7 @@
     border: none;
     cursor: pointer;
     display: flex;
-    height: 60px;
+    height: $header-height;
     justify-content: center;
     left: 0;
     overflow: hidden;
@@ -32,7 +33,7 @@
     width: 60px;
     z-index: 1;
 
-    @media (min-width: $breakpoint-md) {
+    @include isHeaderDesktop {
       display: none;
     }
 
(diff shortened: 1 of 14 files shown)


src/components/Header/_header.scss
```diff
@@ -1,17 +1,17 @@
...
-  padding: 0 $spacing-md;
-  position: fixed;
-  top: 30px;
-  width: 100vw;
+  padding: $user-management-height $spacing-md 0; // save space for the user management
+  position: sticky;
+  top: 0;
 
-  @media (max-width: #{$breakpoint-md - 1}) {
+  @include isHeaderMobile {
```
adietrich3074, line 13: I love this. So much more readable at a glance 👍

adietrich3074 (commented):
Everything looks good for the most part. I'm curious about the after image for the mobile navigation. There's a lot of space between each of those items and I haven't seen any Figma where the mobile nav looks like that. The background is also entirely gray which I have seen on the current mobile nav in prod but I haven't seen it look that way in any of the redesigns of it. Is this the intended final look of the mobile nav?

After looking at the netlify link I'm not seeing all that extra space between menu items but I am seeing the background of the menu all grey and it also looks like the overlay when the menu is open is removed Was that intentional?

The netlify deploy preview also seems to be missing the centering of the Browse All Upcoming link and the label change. Can we clarify which is the correct label and if it should be centered or not?

src/components/Header/_header.scss
scottdickerson: yeah the gray background is there in the newer Figmas.  Great catch on the missing overlay, I had caused this with the wrong `isHeader*` mixin.  I've updated the screenshot at the top to match.

scottdickerson: here's the updated Figma: https://www.figma.com/design/xMuOXOAKVt5HC7hgYjF3ot/Components-v2.0?node-id=7954-18333&t=JnDVzGnoZFwGuUW8-4. As far as the centering of the Browse All Upcoming, since that's content provided by the consumer, I couldn't think of a good way to make that change in seldon itself so only do it in Remix directly 🤷

adietrich3074: Ahh ok I see what you're saying. We make tweaks to it directly in remix. I think this is fine then

adietrich3074 (approved):
LGTM 👍

PR #242, opened 2024-07-24
fix(linklist): L3-2973 relax the linkblock child requirement 

scottdickerson:
**Jira ticket**

[L3-2973](https://phillipsauctions.atlassian.net/browse/L3-2973)

**Summary**

The check in `LinkList` for `LinkBlock` children was too restrictive because it wouldn't allow me to pass a custom component that I wrapped around a `LinkBlock` component.  It would skip rendering my `LinkBlockLocalized` component entirely so I've removed the check.  

- change(LinkList): render any child component, not just `LinkBlock`

- Validate that any child component can render by adding it to a story

**Regression Test**

- Sniff test the existing LinkList story.

**Evidence of testing**

image
image

New Components

[L3-2973]: https://phillipsauctions.atlassian.net/browse/L3-2973?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/LinkList/LinkList.test.tsx
+++ b/src/components/LinkList/LinkList.test.tsx
@@ -1,7 +1,6 @@
 import LinkBlock, { LinkBlockProps } from '../LinkBlock/LinkBlock';
 import { render, screen } from '@testing-library/react';
 import LinkList from './LinkList';
-import Button from '../Button/Button';
 import { runCommonTests } from '../../utils/testUtils';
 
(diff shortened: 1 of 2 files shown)


adietrich3074 (approved):
LGTM 👍 Glad to see this. I had thought to try to use this component for something I was doing in the header but, after realizing it only took LinkBlock children, I decided against it. Maybe we can look into leveraging this when we implement the new header changes

PR #243, opened 2024-07-25
fix(header): L3-2945 language overlay should disappear correctly

scottdickerson:
**Jira ticket**

[L3-2945](https://phillipsauctions.atlassian.net/browse/L3-2945)

**Screenshots**
| Before | After |
| ------ | ----- |
| https://github.com/user-attachments/assets/3e6e8ebe-e2b1-4298-be29-9d4c6bb6fec1 | https://github.com/user-attachments/assets/efa9ef4b-9a5c-4bac-9fcd-89e8b71cb880 |

**Summary**

See the videos above but after selecting a language, the language switcher stuck around

- change(usermanagement): remove focus so the language switcher will disappear after leaving

- In the header story select Chinese as the language and then mouse off the language switcher.  It should totally disappear
- Check the UserManagement story too

**Regression Test**

- Make sure the onLanguageChange gets called in the console.log when switching

**Evidence of testing**

- See after video above and here's the console log:
image

New Components

[L3-2945]: https://phillipsauctions.atlassian.net/browse/L3-2945?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Header/Header.stories.tsx
+++ b/src/components/Header/Header.stories.tsx
@@ -214,7 +214,7 @@ export const Playground = (props: HeaderProps) => (
               <NavigationItem navGroup="nav-link-sm" navType={LinkVariants.navLinkSm} href="#" label="Careers" />
             </NavigationList>
           </NavigationItemTrigger>
-          <UserManagement />
+          <UserManagement onLanguageChange={(language) => console.log('languageChange', language)} />
         </NavigationList>
       </Navigation>
     </Header>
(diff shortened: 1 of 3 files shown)


adietrich3074 (approved):
LGTM 👍

PR #257, opened 2024-08-05
fix(css): L3-2387 rename scss files and refactor into separate files so the partials can be imported separately

scottdickerson:
BREAKING CHANGE: Rename the scss files.  You need to import these two files to get the componentStyles and access to the SASS Partials exported from seldon.
In one global file (should only be imported once in your application)
```
@use '../../node_modules/@phillips/seldon/dist/scss/componentStyles';
```
In each component scss that needs access to the seldon partials:
```
// Seldon partials
@use '../../node_modules/@phillips/seldon/dist/scss/allPartials';
```

**Jira ticket**

[L3-2387](https://phillipsauctions.atlassian.net/browse/L3-2387)

**Summary**

This change was made because we needed consumers to be able to import the seldon component styles once globally, but allow them to import the partials as many times as they need in each component.  This will stop consumers from getting multiple sets of duplicate css classes when they import the partials.

Inside the seldon repo
`npm run build && npm link`

From `phillips-public-remix`
`npm link @phillips/seldon` 

Now import the `allPartials` and `componentStyles` separately.
You should see that the number of duplicate classes are greatly reduced.  A few still remain but it should be much better than before.

**Regression Test**

- Verify that the component still apply their styles correctly when viewing the `/about` page
- Storybook should still render the component styles correctly
- Stories that use the partials tokens should also work correctly (for instance the `color-tokens` story)
- Running `npm run create-component` should create the new component with partials applied

**Evidence of testing**

Only one class applied to the seldon component:
image

New Components

[L3-2387]: https://phillipsauctions.atlassian.net/browse/L3-2387?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


constantinehuzenko (approved):
LGTM

davidicus (approved):

scottdickerson:
I validated this works when linking from `phillips-public` following the instructions in the README.

PR #260, opened 2024-08-05
feat(htmlparser): L3-3194 - create a safe parser for rich text

davidicus:
**Jira ticket**

[L3-3194](https://phillipsauctions.atlassian.net/browse/L3-3194)

**Summary**

This component will sanitize and transform HTML strings we get from the backend and render them using our own Seldon components

- Changed the Link component to have an index page like other components.
- moved Link variant enum to a types.ts file
- Created an HTMLParser component with all the needed files.
- Added export to main index.ts file
- add HTMLParser scss file to main sass file

- View the story where we are passing in the html string. If you open up the inspector you can see the classes from our text and link component are present, showing off the transform that is being performed. 

**Regression Test**

- Link component and story should all work as before. 

**Evidence of testing**

image

New Components

[L3-3194]: https://phillipsauctions.atlassian.net/browse/L3-3194?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


adietrich3074 (approved):
This is an interesting approach. I thought the component would just be parsing strings with html that needed to be interpreted, but I like that this goes further and converts common tags to leverage our Text and Link components with the ability to add more tags as we need them. We'll have to remember to use this where we can as it comes up. LGTM 👍

src/components/HTMLParser/utils.tsx
```diff
@@ -0,0 +1,67 @@
...
+    if (name === 'p') {
+      return (
+        <Text variant={TextVariants.body1}>{domToReact(children as DOMNode[], options as HTMLReactParserOptions)}</Text>
+      );
+    }
+
+    if (name === 'a') {
+      // why do I have to do this?
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      const { class: className, ...rest } = attribs;
```
davidicus, line 56: Anyone know why typescript insisted that I rename the class variable here?

scottdickerson: @davidicus because you're not using the `className` variable anywhere. You can rename it to `_className` to indicate that you intended to strip it out (this pattern is in our eslint-rule) of the destructured attributes if that's what you really want.

PR #264, opened 2024-08-07
feat(subscribe): L4-3126 changed styles

constantinehuzenko:
**Jira ticket**

[L3-3126](https://phillipsauctions.atlassian.net/browse/L3-3126)

**Screenshots**
| Before | After |
| ------ | ----- |
| image | image |

**Summary**

Changed styles to follow new Figma's design

- feat(_subscribe.scss): added new styles
- feat(Subscribe.tsx): added privacy text

- verify that subscribe states are working
- verify that privacy text is visible
- verify that signup button is full width in mobile view

**Regression Test**

**Evidence of testing**

- Post logs, screenshots, etc

New Components

[L3-3126]: https://phillipsauctions.atlassian.net/browse/L3-3126?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Input/_input.scss
+++ b/src/components/Input/_input.scss
@@ -43,7 +43,6 @@ $lg: #{$px}-input--lg;
     -webkit-line-clamp: 2;
     line-height: 1.25;
     overflow: hidden;
-    padding-left: 0;
   }
 
   // Small input size
--- a/src/components/Input/_input.scss
+++ b/src/components/Input/_input.scss
@@ -15,11 +15,10 @@ $lg: #{$px}-input--lg;
   }
 
   &__label {
-    color: $keyline-gray;
-    font-size: 0.8125rem;
+    color: $pure-black;
+    font-size: 0.75rem;
     font-weight: 600;
     margin-bottom: 0.5rem;
-    text-transform: uppercase;
     word-break: break-word;
 
     &--hidden {
@@ -44,6 +43,7 @@ $lg: #{$px}-input--lg;
     -webkit-line-clamp: 2;
     line-height: 1.25;
     overflow: hidden;
+    padding-left: 0;
   }
 
   // Small input size
@@ -127,7 +127,6 @@ $lg: #{$px}-input--lg;
   .#{$px}-input__validation {
     animation: reveal 0.45s linear forwards;
     padding-bottom: 0.5rem;
-    padding-left: 2rem;
   }
 }
 
--- a/src/components/Input/_input.scss
+++ b/src/components/Input/_input.scss
@@ -104,6 +104,10 @@ $lg: #{$px}-input--lg;
   &--invalid {
     color: $error-red;
 
+    label {
+      color: $error-red;
+    }
+
     .#{$px}-input__input,
     .#{$px}-input__input::placeholder,
     .#{$px}-input__validation {
--- a/src/components/Subscribe/Subscribe.tsx
+++ b/src/components/Subscribe/Subscribe.tsx
@@ -52,6 +52,10 @@ export interface SubscribeProps extends React.HTMLAttributes<HTMLFormElement> {
    */
   successText?: string;
   /**
+   * Subscribe privacy text
+   */
+  privacyText?: string;
+  /**
    * Subscribe state for loading or error
    */
   subscriptionState?: SubscriptionState;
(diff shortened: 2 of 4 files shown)


src/components/Subscribe/_subscribe.scss
```diff
@@ -4,34 +4,57 @@
...
+    &__label {
+      color: $pure-black;
+      font-size: 0.75rem;
+      text-transform: none;
+    }
+
+    &__validation {
+      padding-left: 0;
+    }
+  }
```
davidicus, line 20: This should probably be applied to the input component itself. I see that the padding got broken on that input at some point,

src/components/Subscribe/_subscribe.scss
```diff
@@ -4,34 +4,57 @@
...
 .#{$px}-subscribe {
-  max-width: 30rem;
+  max-width: 600px;
   width: 100%;
 
+  .#{$px}-input {
+    &__label {
+      color: $pure-black;
+      font-size: 0.75rem;
+      text-transform: none;
```
davidicus, line 14: This should probably be applied to the input component. From that [figma](https://www.figma.com/design/hMu9IWH5N3KamJy8tLFdyV/EASEL-Compendium%3A-Tokens%2C-Components-%26-Patterns?node-id=7584-17668&m=dev) it looks like this applies to all inputs.

davidicus: Should the font size come from a text mixin?

src/components/Subscribe/Subscribe.tsx
```diff
@@ -51,6 +51,10 @@ export interface SubscribeProps extends React.HTMLAttributes<HTMLFormElement> {
    * Subscribe success text
    */
   successText?: string;
+  /**
+   * Subscribe privacy text
+   */
+  privacyText?: string;
```
davidicus, line 57: are we going to have this privacy text in all areas where we use this component? We are making it conditional in the types but we are always rendering a p tag. We should conditionally render this element.

src/components/Subscribe/Subscribe.tsx
davidicus, line 1: while we are here we should change the jumping that happens when we have error text by rendering a `&nbsp; if no error or warning is present.

scottdickerson (approved):
LGTM

src/components/Subscribe/_subscribe.scss
```diff
@@ -4,34 +4,57 @@
...
+    color: $pure-black;
   }
 
   &__input__wrapper {
     margin-top: $spacing-md;
   }
 
   &__button {
-    margin: $spacing-md 0 $spacing-md;
+    height: 36px;
```
davidicus, line 39: I am not a fan of setting explicit heights on things. The padding, font size, and line-height should equal the height of the button. Adding up the values they have in [this figma](https://www.figma.com/design/koazePmhaPVdUszim1NGTP/Navigation?node-id=2210-1384&m=dev) does not add up to what is listed.

src/components/Subscribe/_subscribe.scss
```diff
@@ -4,34 +4,57 @@
...
+    @include media($size-sm) {
+      width: 240px;
+    }
+  }
+
+  &__privacy {
+    @include text($body3);
+
+    // margin-top: $spacing-sm;
+    background-color: yellow;
```
davidicus, line 52: Where in the figma are you seeing this background color applied. I can not find it.

davidicus (changes requested):
A few small comments here

src/components/Input/_input.scss
```diff
@@ -44,6 +43,7 @@ $lg: #{$px}-input--lg;
     -webkit-line-clamp: 2;
     line-height: 1.25;
     overflow: hidden;
+    padding-left: 0;
```
davidicus, line 46: I think this is not needed with the removal on line 130

davidicus (commented):
This all looks good. I do think the label of the input is supposed to be red  when in an error state. Currently, it is black.

davidicus (approved):

PR #267, opened 2024-08-09
feat(usermanagement)!: L3-2972 add a loading auth state

scottdickerson:
BREAKING CHANGE: The  `UserManagement` `isLoggedIn` property has been replaced with the `authState` property

**Jira ticket**

[L3-2972](https://phillipsauctions.atlassian.net/browse/L3-2972)

**Screenshots**
| Before | After |
| ------ | ----- |
| image | image |

**Summary**

Because of CDN caching requirements, we need for our UserManagement panels LoggedIn/LoggedOut state to be determined client-side only after JavaScript hydrates.  This means we need to add a "Loading" state to the UserManagement panel.

- change(UserManagement): Switch `isLoggedIn` to `authState` enum so more than 2 states can be used
- change(userManagement.scss): fix the items so that their click target doesn't extend past the top area

- Go to the Header and UserManagement stories
- Change the authState property and verify that the section hides while it is loading.

**Regression Test**

- N/A

**Evidence of testing**

- Post logs, screenshots, etc

New Components

[L3-2972]: https://phillipsauctions.atlassian.net/browse/L3-2972?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Header/Header.stories.tsx
+++ b/src/components/Header/Header.stories.tsx
@@ -8,6 +8,7 @@ import { LinkVariants } from '../Link/utils';
 import { px } from '../../utils';
 import UserManagement from '../UserManagement/UserManagement';
 import Search from '../Search/Search';
+import { AuthState } from '../UserManagement/types';
 
 const meta = {
   title: 'Components/Header',
@@ -20,11 +21,15 @@ const meta = {
     },
     layout: 'fullscreen',
   },
-} satisfies Meta<typeof Header>;
+  args: {
+    authState: AuthState.LoggedOut,
+  },
+  argTypes: { authState: { control: { type: 'select' }, options: Object.values(AuthState) } },
+} satisfies Meta<typeof Header & typeof UserManagement>;
 
 export default meta;
 
-export const Playground = (props: HeaderProps) => (
+export const Playground = ({ authState, ...props }: HeaderProps & { authState?: AuthState }) => (
   <div>
     <Header {...props}>
       <Navigation id={`${px}-main-nav`} backBtnLabel="← Back">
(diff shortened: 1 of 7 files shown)


src/components/Header/Header.stories.tsx
```diff
@@ -8,6 +8,7 @@ import { LinkVariants } from '../Link/utils';
 import { px } from '../../utils';
 import UserManagement from '../UserManagement/UserManagement';
 import Search from '../Search/Search';
+import { AuthState } from '../UserManagement/types';
```
aahill50, line 11: Should we prefer to use `import type ...` when importing only types? It helps to avoid some potential issues with importing types, but probably more importantly, `import type` is guaranteed to get removed from any bundled code, whereas regular `import` may not

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export

src/components/UserManagement/UserManagement.tsx
```diff
@@ -49,16 +53,18 @@ const UserManagement = ({
   return (
     <div {...commonProps} className={classnames(baseClassName, className)} {...props}>
       <ul className={`${baseClassName}__account-wrapper`}>
-        {isLoggedIn && (
-          <AccountDetailsComponent>
-            <AccountCircle className={`${baseClassName}__account-icon`} />
-          </AccountDetailsComponent>
+        {authState !== AuthState.Loading && (
```
aahill50, line 56: I always liked pulling these conditional checks out into their own variables. i.e. 
`const showAccountDetails = authState !== AuthState.Loading;`
Not a huge deal here, but it does make things a little easier to maintain down the line if new conditions ever get added.

aahill50 (approved):
Left a couple of minor comments/suggestions, but overall looks good 👍

src/components/Header/Header.stories.tsx
scottdickerson: since this is both a type and a constant I can't do the type import.

src/components/UserManagement/UserManagement.tsx
scottdickerson: good idea!

src/components/Header/Header.stories.tsx
aahill50: Ah, my mistake. I didn't see that it was being used for both... Are we already using `import type` in other places?

PR #276, opened 2024-08-15
fix(accordion): L3-3274 add "variant" prop to Accordion

aahill50:
**Jira ticket**

[L3-3274](https://phillipsauctions.atlassian.net/browse/L3-3274)

**Screenshots**
| Before | After |
| ------ | ----- |
| L3-3274 - before | L3-3274 - after |

**Summary**

Fixes the console error coming from the Accordion component. The error was a little misleading, it was complaining about the `collapsible` prop being passed to the DOM as a boolean instead of a string, but the error was actually caused by the fact that we didn't need to pass in the `collapsible` prop at all because the Accordion `type` was set to `multiple`. When the prop is `multiple`, the child components are already set to be collapsible, and I think that the `collapsible` prop we were passing in was being treated as an extra prop and getting passed all the way through, down to the `div` on the DOM. By removing the extra prop, the component still works the same and it eliminates that error. 

- Accordion.tsx
  - Replaced `type` prop with `variant`
- Accordion/utils.ts
  - Added getAccordionVariantProps helper
- Accordion/types.ts 
  - Added AccordionVariants enum
- Accordion.test.tsx
  - Added tests and increased coverage
- Accordion.stories.tsx
  - Updated to pass in correct props to <Accordion />
- Footer.stories.tsx
  - Updated to pass in correct props to <Accordion />
  
In seldon:
`npm run build`
`npm link`

In phillips-public-remix:
`npm link @phillips/seldon`
`npm run dev`

Navigate to `localhost:3000` and check the remix server for errors. Confirm the below error no longer occurs:
`Warning: Received 'true' for a non-boolean attribute 'collapsible'.`

**Regression Test**

- Follow the steps above and confirm that the Footer component on the page loads and functions properly

**Evidence of testing**

L3-3274 - after 

New Components

[L3-3274]: https://phillipsauctions.atlassian.net/browse/L3-3274?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Accordion/Accordion.tsx
+++ b/src/components/Accordion/Accordion.tsx
@@ -1,41 +1,10 @@
-import { AccordionVariantKey, AccordionVariants } from './types';
+import { AccordionVariantKey } from './types';
 
 import React from 'react';
 import classnames from 'classnames';
 import { getCommonProps } from '../../utils';
 import * as Accordion from '@radix-ui/react-accordion';
-
-interface AccordionVariantProps {
-  /**
-   * Determines whether multiple elements can be opened at the same time or not.
-   */
-  type: 'single' | 'multiple';
-  /**
-   * Determines if an open element can be closed by clicking on it.
-   * Only applicable to the `single` variants.
-   */
-  collapsible?: boolean;
-}
-
-/**
- * Sets the type and collapsible props based on the variant prop
- * The collapsible prop should only be passed when the variant is single
- */
-const getAccordionVariantProps = (variant?: AccordionVariantKey): AccordionVariantProps => {
-  const variantProps: AccordionVariantProps = { type: 'multiple' };
-
-  if (variant === AccordionVariants.singleCollapsible) {
-    variantProps.type = 'single';
-    variantProps.collapsible = true;
-  }
-
-  if (variant === AccordionVariants.singleNonCollapsible) {
-    variantProps.type = 'single';
-    variantProps.collapsible = false;
-  }
-
-  return variantProps;
-};
+import { getAccordionVariantProps } from './utils';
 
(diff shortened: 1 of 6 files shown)


aahill50:
I updated this PR to add a `variant` prop to Accordion instead of just removing the `collapsible` prop that was causing the issue. Now, the component accepts a `variant` prop and dynamically determines which `type` and `collapsible` props to pass to the underlying radix Accordion component

src/components/Accordion/Accordion.tsx
```diff
@@ -1,14 +1,54 @@
...
+}
+
+/**
+ * Sets the type and collapsible props based on the variant prop
+ * The collapsible prop should only be passed when the variant is single
+ */
+const getAccordionVariantProps = (variant?: AccordionVariantKey): AccordionVariantProps => {
+  const variantProps: AccordionVariantProps = { type: 'multiple' };
+
+  if (variant === AccordionVariants.singleCollapsible) {
```
scottdickerson, line 27: I would be tempted to do this as a switch statement in a standalone function, not sure if you agree

scottdickerson (approved):
LGTM

src/components/Accordion/Accordion.tsx
scottdickerson: or at least extract it to a separately testable function and write a testcase specifically targeted at it.

scottdickerson (changes requested):
@aahill50 probably should increase test coverage on the new code before merging.  Could you also fill out the rest of the sections of the PR template?

scottdickerson (approved):
LGTM

PR #280, opened 2024-08-21
feat(dropdown): L3-3342 add new dropdown component

smarks8:
**Jira ticket**

[L3-3342](https://phillipsauctions.atlassian.net/browse/L3-3342)

**Screenshots**
| image | image |

**Summary**

- New Dropdown primitive widget
- Controlled component with no internal state
- Callbacks should be triggered when the value changes
- Styling focus for masthead 

- Adds new Dropdown folder with all component files
- Adds two new svgs (up and down chevron)
- Small update to vite config to ignore index.ts files in coverage report

- Test in storybook link

**Regression Test**

**Evidence of testing**

image

New Components

[L3-3342]: https://phillipsauctions.atlassian.net/browse/L3-3342?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,101 @@
+import React, { useState, ComponentProps } from 'react';
+import { getCommonProps } from '../../utils';
+import classnames from 'classnames';
+import * as DropdownSelect from '@radix-ui/react-select';
+import ChevronDownIcon from '../../assets/chevronDown.svg?react';
+import ChevronUpIcon from '../../assets/chevronUp.svg?react';
```
davidicus, line 6: I would only use one svg and then use css to flip the icon for the different states.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,101 @@
...
+          </DropdownSelect.Portal>
+        </DropdownSelect.Root>
+      </div>
+    );
+  },
+);
+
+Dropdown.displayName = 'Dropdown';
+
+const DropdownItem = React.forwardRef<HTMLDivElement, DropdownSelect.SelectItemProps>(
```
davidicus, line 89: Do we need this to be a forwarded ref? We are not passing any refs to this component. Does Radix do some magic to get this ref passed?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * [Storybook Link](https://phillips-seldon.netlify.app/?path=/docs/components-dropdown--overview)
+ */
+const Dropdown = React.forwardRef<HTMLDivElement, DropdownProps>(
+  ({ options, value, onValueChange, label, className, id, ...props }, ref) => {
+    const { className: baseClassName, ...commonProps } = getCommonProps({ id }, 'Dropdown');
+    const [isOpen, setIsOpen] = useState(false);
+
+    return (
+      <div className={classnames(baseClassName, className)} id={id} {...commonProps} {...props} ref={ref}>
```
davidicus, line 45: Should this ref be passed to the input? Does Radix provide a way to pass along?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,101 @@
...
+        <DropdownSelect.Root
+          defaultValue={value}
+          onValueChange={(value) => {
+            onValueChange(value);
+          }}
+          onOpenChange={setIsOpen}
+        >
+          <DropdownSelect.Trigger className={classnames(`${baseClassName}__trigger`)} aria-label={label}>
```
davidicus, line 53: for any classes that are just the single class and do not change based off conditions we can just pass directly to classNames without the need for the classnames package.

src/components/Dropdown/_dropdown.scss
```diff
@@ -0,0 +1,78 @@
...
+      height: 16px;
+      width: 16px;
+    }
+  }
+
+  &__viewport {
+    display: flex;
+    flex-direction: column;
+    gap: $spacing-xsm;
+    padding: var(--spacing-xsm) 0;
```
davidicus, line 55: We can just use our spacing token here instead of the css variable?

davidicus (changes requested):
a few comments.

src/components/Dropdown/Dropdown.tsx
smarks8: Ah yeah, looks like this ref should be on the trigger according to the docs, will move it

src/components/Dropdown/Dropdown.tsx
smarks8: The main implementation examples had the `DropdownItem` set up like this, so I followed their lead: https://www.radix-ui.com/primitives/docs/components/select

I can remove for now if we want, but I'm not positive of what magic is around it

davidicus: Since focus is fully managed they probably use this ref internally.

src/components/Dropdown/_dropdown.scss
```diff
@@ -0,0 +1,82 @@
...
+.#{$px}-dropdown {
+  @include text($body3);
+
+  &__trigger {
+    align-items: center;
+    background-color: $white;
+    border: none;
+    display: inline-flex;
+    gap: 10px;
+    height: 24px;
```
davidicus, line 12: I don't think we should have an explicit height. ideally the text size and styles along with the padding would equal the height we have in figma. Looking at figma I am also seeing a height of 32px.
image

src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -0,0 +1,27 @@
...
+  { label: '中文', value: SupportedLanguages.zh },
+];
+
+export const Playground = (props: DropdownProps) => {
+  const [defaultValue, setDefaultValue] = useState('en');
+  return <Dropdown {...props} options={languages} value={defaultValue} onValueChange={setDefaultValue} id="test" />;
+};
+
+// More on writing stories with args: https://storybook.js.org/docs/react/writing-stories/args
+Playground.args = {};
```
davidicus, line 25: We should set up the args with the values you have in playground so they can be manipulated in the story. I would like to add more to see what this component looks like with a whole list of items. 
```suggestion
Playground.args = {
    options: language,
    value: 'en'
};
```

src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+export default meta;
+
+const languages = [
+  { label: 'English', value: SupportedLanguages.en },
+  { label: '中文', value: SupportedLanguages.zh },
+];
+
+export const Playground = (props: DropdownProps) => {
+  const [defaultValue, setDefaultValue] = useState('en');
```
davidicus, line 20: Since we are emulating a controlled input here we can just use that value prop as the default value for this state.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,105 @@
...
+          defaultValue={value}
+          onValueChange={(value) => {
+            onValueChange(value);
+          }}
+          onOpenChange={setIsOpen}
+        >
+          <DropdownSelect.Trigger className={`${baseClassName}__trigger`} aria-label={label} ref={ref}>
+            <DropdownSelect.Value placeholder={value} />
+            <DropdownSelect.Icon>
+              {<ChevronDownIcon className={isOpen ? `${baseClassName}__trigger-icon-expanded` : ''} />}
```
davidicus, line 58: This would be the scenario where we would want to use the classnames packages. 
```suggestion
              {<ChevronDownIcon className={classnames({ [`${baseClassName}__trigger-icon-expanded`] : isOpen })} />}
```

src/components/Dropdown/_dropdown.scss
```diff
@@ -0,0 +1,82 @@
...
+      box-shadow: 0 0 0 1px $light-gray;
+    }
+  }
+
+  &__content {
+    background-color: $white;
+    border: 1px solid $light-gray;
+
+    // following will add additional arrows if list is near edge of viewport
+    max-height: var(--radix-select-content-available-height);
```
davidicus, line 39: This is defined by radix library?

davidicus (changes requested):
A few more comments

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,105 @@
...
+  ({ options, value, onValueChange, label, className, id, ...props }, ref) => {
+    const { className: baseClassName, ...commonProps } = getCommonProps({ id }, 'Dropdown');
+    const [isOpen, setIsOpen] = useState(false);
+
+    return (
+      <div className={classnames(baseClassName, className)} id={id} {...commonProps} {...props}>
+        <DropdownSelect.Root
+          defaultValue={value}
+          onValueChange={(value) => {
+            onValueChange(value);
```
constantinehuzenko, line 51: can we move this to separate function?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,105 @@
...
+            </DropdownSelect.Icon>
+          </DropdownSelect.Trigger>
+          <DropdownSelect.Portal>
+            <DropdownSelect.Content className={`${baseClassName}__content`} position="popper">
+              <DropdownSelect.ScrollUpButton className={`${baseClassName}__scroll-button`}>
+                <ChevronUpIcon />
+              </DropdownSelect.ScrollUpButton>
+              <DropdownSelect.Viewport className={`${baseClassName}__viewport`}>
+                {options.map((option) => {
+                  return (
```
constantinehuzenko, line 68: we don't need this return, you can just wrap your component with `()`

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,105 @@
...
+      </div>
+    );
+  },
+);
+
+Dropdown.displayName = 'Dropdown';
+
+const DropdownItem = React.forwardRef<HTMLDivElement, DropdownSelect.SelectItemProps>(
+  ({ children, className, ...props }, forwardedRef) => {
+    return (
```
constantinehuzenko, line 95: we can remove `return` as well

src/components/Dropdown/Dropdown.tsx
scottdickerson: @davidicus this was my idea just in case some consumer needed to measure the width or really call any Javascript imperative function on the element. (is the element in the viewport, etc)

src/components/Dropdown/_dropdown.scss
constantinehuzenko: not sure that we can use this variable

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+export interface DropdownProps
+  extends Omit<DropdownSelect.SelectProps, 'defaultValue' | 'dir'>,
+    Omit<ComponentProps<'div'>, 'ref'> {
+  /**
+   * All options to be listed in the dropdown
+   */
+  options: {
+    label: string;
+    value: string;
```
scottdickerson, line 16: could we export this as it's own typescript type for consumers to use?  `export interface DropdownItem { label: string; value: string } `

src/components/Dropdown/_dropdown.scss
```diff
@@ -0,0 +1,82 @@
...
+  &__item {
+    @include text($body3);
+
+    align-items: center;
+    display: flex;
+    padding: 0 $padding-sm;
+    position: relative;
+    user-select: none;
+
+    &[data-disabled] {
```
scottdickerson, line 71: I've always really liked this approach @smarks8

scottdickerson:
I find this disabled state pretty hard to read: 
image

Could we make it a bit darker?

scottdickerson:
> I find this disabled state pretty hard to read: image
> 
> Could we make it a bit darker?

TBH, do we really need to change the style?  The fact that it's not clickable should communicate that well I think

src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -0,0 +1,31 @@
...
+  { label: '中文', value: SupportedLanguages.zh },
+  { label: 'Español', value: 'es' },
+  { label: 'Français', value: 'fr' },
+  { label: 'Deutsch', value: 'de' },
+];
+
+export const Playground = (props: DropdownProps) => {
+  const [value, setValue] = useState('en');
+  return <Dropdown {...props} options={languages} value={value} onValueChange={setValue} id="test" />;
+};
```
davidicus, line 25: You will need to make these changes for the story controls to have any affect.. 

```suggestion
export const Playground = (props: DropdownProps) => {
  const [value, setValue] = useState(props.value);
  return <Dropdown {...props} options={props.options} value={value} onValueChange={setValue} id="test" />;
};
```

scottdickerson (approved):
LGTM

src/components/Dropdown/Dropdown.tsx
davidicus: Yeah if that is the case we have to be able to pass the ref from props, no?

PR #284, opened 2024-08-21
feat(footer): L3-3126 3rd footer update

constantinehuzenko:
**Jira ticket**

[Jira Ticket ID](https://phillipsauctions.atlassian.net/browse/L3-3126)

**Screenshots**
| Before | After |
| ------ | ----- |
| image | image |

**Summary**

Added new footer update

- feat(Footer.tsx): removed split panel
- feat(_footer.scss): removed unnecessary styles
- feat(Social.tsx): changed text tag

- verify that desktop version is not in accordion
- verify that mobile version is in accordion
- verify that links containers have the same width on desktop
- verify that tablet view has max-width for footer content

**Regression Test**

**Evidence of testing**

- Post logs, screenshots, etc

New Components

### DIFF
--- a/src/components/Footer/Footer.stories.tsx
+++ b/src/components/Footer/Footer.stories.tsx
@@ -19,6 +19,84 @@ const meta = {
 
 export default meta;
 
+export const Playground = (props: FooterProps) => (
+  <Footer {...props}>
+    {/* desktop version of footer */}
+    <div className={`${px}-footer-desktop`}>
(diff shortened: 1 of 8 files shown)


davidicus (commented):
Looks good. One comment on explicitly setting the width. 

On really large screens we will have some pretty spread out UI. Just an observation. Just wondering if we put a max width on our content even if it is really large, just to prevent this case?
image

davidicus (approved):

PR #286, opened 2024-08-23
feat(header-ux-updates): L3-3366 header ux updates search

adietrich3074:
**Jira ticket**

[L3-3366](https://phillipsauctions.atlassian.net/browse/L3-3366)

**Screenshots**
| Before | After |
| ------ | ----- |
| Screenshot 2024-08-23 112403 | Screenshot 2024-08-23 112539 |

| Mobile |
Screenshot 2024-08-23 112558
Screenshot 2024-08-23 112652

**Summary**

Updated the Search component to match styles and functionality. Updated unit tests to match the new functionality

- Style adjustments
- Updated unit tests

- Verify in the Search story of our Storybook

**Evidence of testing**

- Screenshot 2024-08-23 114816

New Components

[L3-3366]: https://phillipsauctions.atlassian.net/browse/L3-3366?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


constantinehuzenko (commented):
when your tab is focused on the link, esc to close doesn't work

image

constantinehuzenko (approved):
really great work, couple minor changes but not necessary

adietrich3074:
> when your tab is focused on the link, esc to close doesn't work
> 
> image

I forwarded the keydown event to the SearchResults component so this should work now

adietrich3074:
> an underline when mousing over the Search icon seems strange to me: image since it's more of a button than a link

@scottdickerson This is how it reacts when hovered in the protopie
https://cloud.protopie.io/p/1b0a7def5ee11e505e4808b4

src/components/Search/SearchResults/SearchResults.tsx
```diff
@@ -7,30 +7,35 @@ export interface SearchResultsProps extends React.HTMLAttributes<HTMLElement> {
   isLoading?: boolean;
```
scottdickerson, line 7: Can we do this so the results interface is sharable with consumers and then reuse this interface in Search.stories.

```
export interface SearchResult {
  id: string;
  url: string;
  label: string;
}

...
autoCompleteResults?: SearchResult;
```

src/components/Search/Search.tsx
```diff
@@ -53,30 +51,39 @@ export interface SearchProps extends React.HTMLAttributes<HTMLDivElement> {
...
   defaultValue = '',
   className,
-  placeholder = 'Type to search',
+  placeholder,
   searchButtonText = 'Search',
-  loadingText = 'Loading...',
+  loadingText = 'Search In Progress...',
   invalidText = 'Invalid search',
   getAllResultsText = (searchValue) => `View all results for ${searchValue}`,
   getAllResultsLink = (searchValue) => `/Search?Search=${searchValue}`,
```
scottdickerson, line 64: I'm seeing some strange changes to the searchValue when we search on it in dev.phillips.com.  I searched for `test test test` and the URL was: https://www.dev.phillips.com/Search?Search=test+test+test+test

Is there something we need to do to the `searchValue` with this default `getAllResultsLink` function?

src/components/Search/Search.tsx
```diff
@@ -53,30 +51,39 @@ export interface SearchProps extends React.HTMLAttributes<HTMLDivElement> {
...
+      return;
+    }
+    searchInputRef.current?.focus();
+  };
+
+  const onKeyDown = (e: React.KeyboardEvent<HTMLInputElement> | React.KeyboardEvent<HTMLAnchorElement>) => {
+    if (e.key === 'Escape') {
+      setSearchEnabled(false);
+      searchFormRef.current?.reset();
     }
```
scottdickerson, line 87: when I hit `Enter` in the search input it doesn't seem to take me to the `Search?Search=value` like cliking the View All Details link does

src/components/Search/Search.tsx
adietrich3074: I think they are doing a weird text replace that I would not like to repeat with our search. Right now I just encode the special charaters normally. When the actual search for the maker hits the backend I believe all special characters are removed anyway so I don't think we have a problem here

adietrich3074: If you replace test+test+test+test with test20%test20%test20%test it returns with the same result set so I think it should be fine to use the encoded value instead of doing our own string replace

scottdickerson (approved):
LGTM

PR #287, opened 2024-08-26
fix(accordion): L3-3439 accordion fixes

aahill50:
**Jira ticket**

[L3-3439](https://phillipsauctions.atlassian.net/browse/L3-3439)

**Screenshots**
| Before | After |
| ------ | ----- |
| https://github.com/user-attachments/assets/49ae04b5-010a-4ca5-bfab-c3458085baca | https://github.com/user-attachments/assets/dd5fa1eb-d45f-4f57-b334-f535fa9d00ee |

**Summary**

I updated the Accordion component to fix the behavior of the icons when sections are expanded and collapsed. Previously, they were relying on custom logic to determine the "collapsed" state of each panel, but there is an attribute that radix already uses to determine the collpased state, `data-state="open"` and `data-state="closed"`. I just rendered the icons in the component and use css and the `datta-state` to determine which one to hide and which one to show.

* src/components/Accordion/Accordion.test.tsx
  * Updated tests
* src/components/Accordion/AccordionItem.tsx
  * Updated logic used to show open/close icons
* src/components/Accordion/_accordion.scss
  * Updated css to support new logic
* src/components/Accordion/types.ts
  * Removing typing of extra props that are no longer needed
* src/components/Accordion/utils.ts
  * Added a helper: `getIconClasses` to DRY up AccordionItem

Verify that any uses of the @seldon Accordion component that are using a `single*` variant have panels that open and close properly and the icons on those panels update accordingly based on whether the panel is open or closed.

**Regression Test**

**Evidence of testing**

See attached video

New Components

[L3-3439]: https://phillipsauctions.atlassian.net/browse/L3-3439?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Accordion/Accordion.test.tsx
+++ b/src/components/Accordion/Accordion.test.tsx
@@ -91,12 +91,20 @@ describe('Accordion', () => {
         ))}
       </Accordion>,
     );
-    expect(screen.queryByTestId(/accordion-item-0-plusIcon/)).toBeInTheDocument();
-    expect(screen.queryByTestId(/accordion-item-1-lockedIcon/)).toBeInTheDocument();
+
+    // Should have 2 closed items
+    expect(container.querySelectorAll(`.phillips-accordion-item[data-state="open"]`)).toHaveLength(0);
+    expect(container.querySelectorAll(`.phillips-accordion-item[data-state="closed"]`)).toHaveLength(2);
+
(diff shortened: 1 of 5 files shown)


scottdickerson (approved):
LGTM

PR #298, opened 2024-09-09
feat(Pagination): L3-2774 Build lot pagination component in Seldon library

zhenpanda:
**Jira ticket**

[L3-2774](https://phillipsauctions.atlassian.net/browse/L3-2774)

**Screenshots**

https://github.com/user-attachments/assets/78427664-9fbb-4fba-9a81-1acb9d3985f4

**Summary**

A pagination component using `<Button>` + `<Select>` with a custom style variant of "inline"

- New Collapsible and ReadMore components

- Pull down branch
- Run storybook
- Check out ReadMore & Collapsible components

**Regression Test**

**Evidence of testing**

- Post logs, screenshots, etc

New Components

[L3-2774]: https://phillipsauctions.atlassian.net/browse/L3-2774?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,140 @@
...
+  React.useEffect(() => {
+    selectedValue && onChange && onChange(selectedValue);
+  }, [onChange, selectedValue]);
+
+  return (
+    <div
+      className={classnames(`${px}-${type}`, { [`${baseClassName}__wrapper`]: baseClassName }, className)}
+      {...commonProps}
+      id={props?.id}
+      onClick={onClick}
```
sgrund14, line 103: should this maybe be separated into `onPrevClick` and `onNextClick` props that are passed to the `Button` components (which could then be called by the locally defined `onPrevClick` and `onNextClick` functions)

having an `onClick` on the wrapper could be confusing

src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+  const paginationClick = (e: React.MouseEvent<HTMLElement>) => {
+    const buttonType = e.currentTarget.dataset.testid;
+    const selectedIndex = selectRef.current?.selectedIndex;
+    const firstIndexValue = selectRef.current?.options[0]?.value;
+    const lastIndex = selectRef.current?.childElementCount ? selectRef.current?.childElementCount - 1 : 0;
+    const lastIndexValue = selectRef.current?.options[lastIndex]?.value;
+    if (selectedIndex !== undefined && firstIndexValue && lastIndexValue) {
+      const prevValue = selectRef.current?.options[selectedIndex - 1]?.value;
+      const nextValue = selectRef.current?.options[selectedIndex + 1]?.value;
```
sgrund14, line 82: does this have the wrapping logic?

when you are on last item, clicking next should bring you to first item
when you are on first item, clicking prev should bring you to last item

src/components/Pagination/Pagination.stories.tsx
```diff
@@ -0,0 +1,60 @@
...
+    control: {
+      type: 'text',
+    },
+  },
+};
+
+export const Playground = ({ playgroundWidth, ...args }: StoryProps) => (
+  <div style={{ width: playgroundWidth, margin: '1rem' }}>
+    <Pagination key={args.value} {...args} id="Pagination-1">
+      <option>Lot 1</option>
```
sgrund14, line 41: is there some way to enforce/document that the children need to be `<option>` here?

part of me wonders if an `options` prop might be simpler, but I guess it's not following the composable pattern as much

what do you think @scottdickerson

src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,140 @@
...
+  children,
+  ...props
+}: PaginationProps) => {
+  const type = 'pagination';
+  const { className: baseClassName, ...commonProps } = getCommonProps(props, 'Pagination');
+  const { id, value, isDisabled } = props;
+  const [selectedValue, setSelectedValue] = React.useState(value);
+  const selectRef = React.useRef<HTMLSelectElement>(null);
+
+  const paginationClick = (e: React.MouseEvent<HTMLElement>) => {
```
sgrund14, line 74: could this be split out into `onPrevClick` and `onNextClick` functions?

might be a little more readable, and require less conditional branching

scottdickerson:
I'm seeing the value control highlighted in Red by the storybook: 
image

scottdickerson:
When I hover over Lot1, I'm not seeing the `cursor: pointer` and there isn't any hover affordance, nor focus style when focused on it with keyboard navigation

src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,140 @@
...
+ *
+ * [Figma Link](https://www.figma.com/design/hxqgsE26wM7hII0WaUaDfF/RW---TIMED-Lot-Details-(PDP)?node-id=1-2&node-type=CANVAS&m=dev)
+ *
+ * [Storybook Link](https://phillips-seldon.netlify.app/?path=/docs/components-pagination--overview)
+ */
+
+const Pagination = ({
+  className,
+  onChange,
+  onClick,
```
scottdickerson, line 62: I think we can get away without an `onClick` and just use the `onChange` callback.

src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,140 @@
+import * as React from 'react';
+import classnames from 'classnames';
+import { getCommonProps } from '../../utils';
+import { px } from '../../utils';
+
+import Select from '../Select/Select';
+import Button from '../Button/Button';
+import chevronRight from '../../assets/chevronRight.svg';
+
+export interface PaginationProps {
```
scottdickerson, line 10: this should extend the underlying element `ComponentProps` being rendered as the root-element.  Then any non-specified props should be passed through

src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,140 @@
...
+  };
+
+  React.useEffect(() => {
+    selectedValue && onChange && onChange(selectedValue);
+  }, [onChange, selectedValue]);
+
+  return (
+    <div
+      className={classnames(`${px}-${type}`, { [`${baseClassName}__wrapper`]: baseClassName }, className)}
+      {...commonProps}
```
scottdickerson, line 101: need to also pass through any other props `{...props}` from above

src/components/Pagination/Pagination.tsx
scottdickerson: I agree, I'd just remove this `onClick` prop entirely

src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,140 @@
...
+  ...props
+}: PaginationProps) => {
+  const type = 'pagination';
+  const { className: baseClassName, ...commonProps } = getCommonProps(props, 'Pagination');
+  const { id, value, isDisabled } = props;
+  const [selectedValue, setSelectedValue] = React.useState(value);
+  const selectRef = React.useRef<HTMLSelectElement>(null);
+
+  const paginationClick = (e: React.MouseEvent<HTMLElement>) => {
+    const buttonType = e.currentTarget.dataset.testid;
```
scottdickerson, line 75: as much as possible, I want to avoid introspecting the DOM element to make decisions.  Instead you can change the callback function to pass the value below.

src/components/Pagination/Pagination.tsx
scottdickerson: I would prefer this: 

```suggestion
  const paginationClick = (newValue:string, e: React.MouseEvent<HTMLElement>) => {
```

src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,140 @@
...
+   */
+  onClick?: (event: React.MouseEvent<HTMLElement>) => void;
+  /**
+   * Optional `onChange` handler that is called whenever the `<Pagination>` value is changed
+   */
+  onChange?: (selectedValue: string) => void;
+  /**
+   * Specify the value of the `<Pagination>`
+   */
+  value?: string | undefined;
```
scottdickerson, line 42: does a `string` or a `number` make more sense?  To me `number` makes the most sense.  I also wonder should this be called `defaultValue` since after the component is rendered once it ignores this `value` prop and instead uses it's internal `setState` (i.e. if it changes it doesn't get updated

src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,140 @@
...
+        nextValue ? setSelectedValue(nextValue) : setSelectedValue(firstIndexValue);
+    }
+  };
+  const paginationSelect = (e: Event) => {
+    const target = e.target as HTMLSelectElement;
+    setSelectedValue(target.value);
+  };
+
+  React.useEffect(() => {
+    selectedValue && onChange && onChange(selectedValue);
```
scottdickerson, line 95: can we remove this `useEffect` and just trigger the `onChange` callback separately in both `paginationSelect` and `paginationClick`  I try to reduce this as much as possible.  (For instance on first render, it's going to trigger this `onChange` even though nothing has changed with the `useEffect` like this.

src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,140 @@
...
+      >
+        {isDisabled || !children ? <option>{disabledText}</option> : children}
+      </Select>
+
+      <Button
+        className={`${px}__pagination-button`}
+        onClick={(e) => paginationClick(e)}
+        data-testid={`${id}-next-button`}
+        isDisabled={isDisabled}
+        aria-label="Next"
```
scottdickerson, line 132: needs to support translating this value

src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,140 @@
...
+      >
+        <img src={chevronRight} className={`${px}-button-icon`} />
```
scottdickerson, line 134: this won't work in consumers unless you import the `chevronRight` as an actual react component and don't use the `<img` tag

src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,140 @@
+import * as React from 'react';
+import classnames from 'classnames';
+import { getCommonProps } from '../../utils';
+import { px } from '../../utils';
+
+import Select from '../Select/Select';
+import Button from '../Button/Button';
+import chevronRight from '../../assets/chevronRight.svg';
```
scottdickerson, line 8:
```suggestion
import chevronRight from '../../assets/chevronRight.svg?react';
```

src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,140 @@
...
+  const { className: baseClassName, ...commonProps } = getCommonProps(props, 'Pagination');
+  const { id, value, isDisabled } = props;
+  const [selectedValue, setSelectedValue] = React.useState(value);
+  const selectRef = React.useRef<HTMLSelectElement>(null);
+
+  const paginationClick = (e: React.MouseEvent<HTMLElement>) => {
+    const buttonType = e.currentTarget.dataset.testid;
+    const selectedIndex = selectRef.current?.selectedIndex;
+    const firstIndexValue = selectRef.current?.options[0]?.value;
+    const lastIndex = selectRef.current?.childElementCount ? selectRef.current?.childElementCount - 1 : 0;
```
scottdickerson, line 78: this logic feels like a separate utility function with an optional "infinite" capability to wrap around, can we extract it to that?

src/components/Pagination/_pagination.scss
```diff
@@ -0,0 +1,64 @@
...
+  display: flex;
+  flex-direction: row;
+}
+
+.#{$px} {
+  &--inline-pagination__wrapper {
+    select {
+      appearance: none;
+      border: none;
+      font-size: $body-size2;
```
scottdickerson, line 13: we should use the `@include text($body-size2)` here instead if possible.

src/components/Pagination/_pagination.scss
```diff
@@ -0,0 +1,64 @@
...
+      font-weight: 600;
+      line-height: 2.5rem;
+      margin-bottom: 0;
+      outline: none;
+      text-align: center;
+      text-align-last: center;
+      text-decoration: underline;
+      text-underline-offset: 10px;
```
scottdickerson, line 21: whoa I didn't know about this!  This would have made my `<Link` components much easier to implement!

src/components/Pagination/_pagination.scss
```diff
@@ -0,0 +1,64 @@
...
+    select {
+      appearance: none;
+      border: none;
+      font-size: $body-size2;
+      font-weight: 600;
+      line-height: 2.5rem;
+      margin-bottom: 0;
+      outline: none;
+      text-align: center;
+      text-align-last: center;
```
scottdickerson, line 19: is this needed?

src/components/Pagination/_pagination.scss
```diff
@@ -0,0 +1,64 @@
...
+    background: none;
+  }
+
+  &:disabled {
+    border: none;
+  }
+}
+
+.#{$px}-left-arrow {
+  transform: scaleX(-1);
```
scottdickerson, line 57: nice!

src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,140 @@
...
+   */
+  id: string;
+  /**
+   * Base class for Pagination component.
+   */
+  className?: string;
+  /**
+   * Child element pass in to display as options to select from.
+   */
+  children?: React.ReactNode;
```
scottdickerson, line 22: I'd rather have this be `pageValues: number[]`, and wrap those in the `<option` before passing as `Select` children

src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,140 @@
...
+  onChange,
+  onClick,
+  variant = 'inline',
+  disabledText = 'Lot',
+  children,
+  ...props
+}: PaginationProps) => {
+  const type = 'pagination';
+  const { className: baseClassName, ...commonProps } = getCommonProps(props, 'Pagination');
+  const { id, value, isDisabled } = props;
```
scottdickerson, line 70: I'd like to remove the `useState` entirely and make this a fully controlled component with no internal state.  I.e. the current `value` is always used, and the `onChange` is called instead of `setSelectedValue`

scottdickerson (approved):
Works great thanks!

PR #307, opened 2024-09-13
feat(seldonprovider): L3-3725 add new seldon provider component that provides media query context

scottdickerson:
**Jira ticket**

[Jira Ticket ID](https://phillipsauctions.atlassian.net/browse/L3-3725)

**Summary**

Some of our components need to render different elements to the DOM based on breakpoints.  This new Provider allows them to do so in a SSR compliant way.

- change(.storybook): lots of prettier formatting, new `CustomStoryWrapper` that makes our `<SeldonProvider` available in every story
- change(Header): use the new SSR Media Query to decide when to show `<LanguageSelector`, I can remove all the custom scss I was doing as well to hide/show it
- change(LanguageSelector): use the new SSR Media Query to decide which version of the `<LanguageSelector` to show, I can also remove the hide/show custom scss there and then simplify the selectors
- change(Navigation): use SSRMediaQuery to decide when to hide/show the language selector
- add(SeldonProvider): central context provider for Seldon, currently just supports the media queries

-  Read the Storybook documentation for the new SeldonProvider component
-  After rendering in storybook you should be able to inspect the DOM and see that the LanguageSelector is only rendered once, not twice.  Moving between mobile and desktop breakpoints will switch the LanguageSelector that's being rendered, but again only one should be rendered.

**Regression Test**

- (Optional) Compare old seldon header to new one and then should look exactly the same at the different breakpoints

**Evidence of testing**

image

New Components


adietrich3074 (approved):
LGTM 👍 The nav item stuff I can use this for is always rendered in the same place so it should be even less complicated than your language selector use case

PR #320, opened 2024-09-17
fix(pagination): L3-2774 use IconButton in Pagination component

scottdickerson:
**Jira ticket**

[Jira Ticket ID](https://phillipsauctions.atlassian.net/browse/L3-2774)

**Summary**

Switch pagination to use IconButton so we have consistent hover states.

- change(Pagination): switch `<Button` to `<IconButton`, and make the `selectLabel` translatable for screenreaders

- Run the pagination story, check the Next/Previous button, change the `selectLabel` prop and in the devTools accessibility inspector for the Select you should see the aria-label is used.

**Regression Test**

- Regress any other pagination functionality

**Evidence of testing**

- Hovered with this state
 
image

New Components


zhenpanda (approved):
lgtm

PR #322, opened 2024-09-18
fix(header-submenus): L3-2818 style fix for view all links in header submenus

adietrich3074:
**Summary**

Last child links in nav list were being styled incorrectly. Added a specific class for view all links to be styled differently than the rest of links in that section

- Added classname view-all to the links that required the specific styling and update the css rules

- Check that the view all links now have the classname 'view-all' applied to them

New Components


src/components/Navigation/NavigationItem/_navigationItem.scss
```diff
@@ -77,18 +77,8 @@
...
-    @include isHeaderMobile {
-      margin-top: $spacing-md;
-    }
-  }
-
-  &--start {
-    margin-right: 1rem;
 
-    & li:last-child {
+    & .view-all {
```
scottdickerson, line 81: hmm, this seems strange to have this as a special case and require a downstream consumer to pass the right class.  Maybe just remove the special styling and let the consumer override for the last link somehow?

scottdickerson (approved):
LGTM

PR #327, opened 2024-09-20
fix(boldedFonts): L3-3693 variable font weight not working in Safari

scottdickerson:
**Jira ticket**

[Jira Ticket ID](https://phillipsauctions.atlassian.net/browse/L3-3693)

**Screenshots**
| Before | After |
| ------ | ----- |
| image | image |

**Summary**

Font-weight > 500 looks incorrect with variable fonts on Safari

- change(*.scss): change `font-weight` to use variable font  `wght` property instead
- change(type-tokens): remove some redundant examples
- change(eslint): put in rule to prevent future `font-weight` props

- Open the "Type Tokens" story on Safari, verify any font-weight: 600 or font-weight: 700 tokens look correct and match Chrome
- Check all the `<Text` component variants comparing Safari and Chrome and they should look the same.
- Check the eslint rule catches a `font-weight: 700;`

**Regression Test**

- Double-check on Chrome, Firefox, and Edge that no regressions

**Evidence of testing**

- See screenshots above

New Components

### DIFF
--- a/src/components/Accordion/_accordion.scss
+++ b/src/components/Accordion/_accordion.scss
@@ -21,7 +21,7 @@
 
   &[data-state='open'] {
     display: block;
-    font-weight: 400;
+    font-variation-settings: 'wght' 400;
     height: auto;
   }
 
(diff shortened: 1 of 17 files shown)


davidicus (approved):
👍

PR #335, opened 2024-09-24
feat(Detail): L3-3155 create Detail component

aahill50:
**Jira ticket**

[L3-3155](https://phillipsauctions.atlassian.net/browse/L3-3155)

**Screenshots**
| Before | After |
| ------ | ----- |
| N/A | image |

**Summary**

Creating a simple `Detail` component, that can be used to display any label/value pairs on the Lot Details page

- Created Detail component

**Regression Test**

**Evidence of testing**

- Post logs, screenshots, etc

New Components

[L3-3155]: https://phillipsauctions.atlassian.net/browse/L3-3155?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Detail/_detail.scss
```diff
@@ -0,0 +1,22 @@
+@use '#scss/allPartials' as *;
+
+.#{$px}-detail {
+  display: flex;
+  justify-content: space-between;
```
scottdickerson, line 5: @aahill50 Rather than a `space-between`, I see a `gap: $spacing-md` in the Figma. 
image
There's also some tricky bits going on where they all align with each other.  Can you create a `DetailList` component that will make sure they all line up on the longest `Detail` and make a story to match this Figma content: https://www.figma.com/design/hMu9IWH5N3KamJy8tLFdyV/EASEL-Compendium%3A-Tokens%2C-Components-%26-Patterns?node-id=11364-24078&m=dev so we make sure it works correctly?

aahill50: @scottdickerson I updated the component, added a DetailsContainer and added some props that determine how the Details are laid out in the container. It supports both the justified layout seen in the Estimate/Starting Bid stack, and the column type layout seen in the Lot Details stack. I also added a prop for showing or hiding the separators since that also differs between the 2 use cases

src/patterns/DetailList/DetailList.tsx
```diff
@@ -0,0 +1,64 @@
...
+   * Whether to render separators between each Detail component
+   */
+  hasSeparators: boolean;
+}
+/**
+ * ## Overview
+ *
+ * A container for displaying a list of [Detail](/docs/components-detail--overview) components. The alignment prop determines whether each Detail's label and value are aligned in columns or justified. Separators can also be added between each Detail component.
+ *
+ * [Figma Link](Add Figma URL here)
```
scottdickerson, line 26: It would be good to point to where this is used in the Figma @aahill50

src/patterns/DetailList/DetailList.tsx
```diff
@@ -0,0 +1,64 @@
...
+          justified: alignment === 'justified',
+        })}
+        {...props}
+        ref={ref}
+      >
+        {children?.map((child, index) => (
+          <div
+            className={classnames(`${baseClassName}-wrapper`, {
+              ['has-separators']: hasSeparators,
+              columns: alignment === 'columns',
```
scottdickerson, line 49: I think it's risky not to prefix the `baseClassName` for these in case other libraries might have used these classnames

src/patterns/DetailList/DetailList.tsx
```diff
@@ -0,0 +1,64 @@
...
+              justified: alignment === 'justified',
+            })}
+            key={`detail-list-wrapper-${index}`}
```
scottdickerson, line 52: using the index here will cause problems if we ever support ways to sort the items after the initial render.  If there's any thing on the child props we could use it might be better

src/patterns/DetailList/DetailList.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { getCommonProps } from '../../utils';
+import classnames from 'classnames';
+import { DetailComponent } from '../../components/Detail';
+
+// You'll need to change the ComponentProps<"htmlelementname"> to match the top-level element of your component
+export interface DetailListProps extends ComponentProps<'div'> {
+  /**
+   * Determines whether each Details' label and value are aligned in columns or justified
+   */
+  alignment: 'columns' | 'justified';
```
scottdickerson, line 11: this is a great idea @aahill50.  It might be good to define a `DetailsListAlignment` enum and export it

src/components/Detail/Detail.tsx
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * [Storybook Link](https://phillips-seldon.netlify.app/?path=/docs/components-detail--overview)
+ *
+ */
+const Detail = forwardRef<HTMLDivElement, DetailProps>(({ className, ...props }, ref) => {
+  const { className: baseClassName, ...commonProps } = getCommonProps(props, 'Detail');
+
+  return (
+    <div {...commonProps} className={classnames(baseClassName, className)} {...props} ref={ref}>
+      <span className={`${baseClassName}--detail-label`}>{props.label}</span>
```
scottdickerson, line 30: does this make it have a double `-detail` in the className?  I think more standard BEM would be `${baseClassName}__label` and `${baseClassName}__value`

src/components/Detail/Detail.tsx
```diff
@@ -0,0 +1,40 @@
...
+      <span className={`${baseClassName}--detail-value`}>{props.value}</span>
```
scottdickerson, line 31: I believe for accessibility it's better to render this tag semantically: https://benmyers.dev/blog/on-the-dl/

src/patterns/DetailList/DetailList.tsx
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * [Storybook Link](https://phillips-seldon.netlify.app/?path=/docs/patterns-detaillist--overview)
+ *
+ */
+const DetailList = forwardRef<HTMLDivElement, DetailListProps>(
+  ({ alignment, className, children, hasSeparators, ...props }, ref) => {
+    const { className: baseClassName, ...commonProps } = getCommonProps(props, 'DetailList');
+
+    return (
+      <div
```
scottdickerson, line 36: I think semantically this might need to be some kind of `<dl` https://benmyers.dev/blog/on-the-dl/

scottdickerson (changes requested):
made some comments, for me the main thing to address is the semantic HTML

src/patterns/DetailList/DetailList.tsx
aahill50: Fixed

src/patterns/DetailList/DetailList.tsx
aahill50: Good point, I added the baseClassNames in

src/patterns/DetailList/DetailList.tsx
aahill50: Fixed this by using the label and value props passed to the child, which I didn't even realize you could from `children`... TIL!

src/patterns/DetailList/DetailList.tsx
aahill50: Done

src/components/Detail/Detail.tsx
aahill50: Yeah, it was making it a double `detail` in the className. Renamed these using `__label` and `__value`, which makes a lot more sense 😅

src/components/Detail/Detail.tsx
aahill50: Cool, I forgot all about the `<dl />` element

src/components/Detail/Detail.test.tsx
```diff
@@ -0,0 +1,13 @@
+import Detail from './Detail';
+import { runCommonTests } from '../../utils/testUtils';
+import { render } from '@testing-library/react';
+
+describe('Detail', () => {
+  runCommonTests(Detail, 'Detail');
+
+  it('should render the label and value', () => {
+    const { getByText } = render(<Detail label="Label" value="Value" />);
+    expect(getByText('Label')).toBeInTheDocument();
```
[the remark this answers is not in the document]
aahill50: Done

src/patterns/DetailList/DetailList.tsx
aahill50: Changed to another `<dl />`

src/components/Detail/Detail.tsx
```diff
@@ -0,0 +1,40 @@
...
+ * [Figma Link](https://www.figma.com/design/hxqgsE26wM7hII0WaUaDfF/RW---TIMED-Lot-Details-(PDP)?node-id=189-5590&node-type=instance&m=dev)
+ *
+ * [Storybook Link](https://phillips-seldon.netlify.app/?path=/docs/components-detail--overview)
+ *
+ */
+const Detail = forwardRef<HTMLDListElement, DetailProps>(({ className, ...props }, ref) => {
+  const { className: baseClassName, ...commonProps } = getCommonProps(props, 'Detail');
+
+  return (
+    <dl {...commonProps} className={classnames(baseClassName, className)} {...props} ref={ref}>
```
scottdickerson, line 29: I think we only want one `dl` at the DetailsList level, not within each Detail @aahill50 .  This can be a `div` to allow you to set those styles as you want and pass the ref

src/components/Detail/Detail.tsx
```diff
@@ -0,0 +1,40 @@
+import { ComponentProps, forwardRef } from 'react';
+import { getCommonProps } from '../../utils';
+import classnames from 'classnames';
+
+export interface DetailProps extends ComponentProps<'dl'> {
```
scottdickerson, line 5: ComponentProps<'div'>

src/components/Detail/Detail.tsx
aahill50: Makes sense

scottdickerson (approved):
Works great thanks!

src/patterns/DetailList/DetailList.tsx
davidicus: We should always be doing BEM class naming so we know its coming from us and belongs to this component.

davidicus (approved):
👍

src/components/Detail/_detail.scss
```diff
@@ -10,8 +11,21 @@
   @include text($string2);
 
   &__label {
+    flex-shrink: 0;
     font-variation-settings: 'wght' 600;
```
scottdickerson, line 15: this is not a blocking thing, but typically we go the other way mobile at the top, then override at the higher breakpoints using the `@include media($size-md) {`

PR #339, opened 2024-09-26
feat(pagination): L3-3826 support options object with translated label

scottdickerson:
**Jira ticket**

[Jira Ticket ID](https://phillipsauctions.atlassian.net/browse/L3-3826)

**Summary**

The `<Pagination` component only took an array of strings or numbers for it's options, and would return that value when the page was changed.  This makes it difficult when translations are involved because sometimes the strings are translated. This feature allows us to pass an object with both a "label" and a "value" as an option so that we can use the "value" as the current state of selection, and "label" to show the translated text.

- change(Pagination): support `PaginationOption` object with label and value and adjust the code so that selections and next previous buttons can handle that and pass back the correct "value" property
- change(_pagination.scss): fix the color
- change(_iconButton.scss): increase size of the svg from 16px to 20px

- check the testcase
- Check the stories both string and object options

**Regression Test**

- Check Modal, Drawer and IconButton stories to make sure the tertiary button within them looks good

**Evidence of testing**

image

New Components


src/components/Pagination/Pagination.stories.tsx
```diff
@@ -32,10 +38,9 @@ const argTypes = {
     control: { type: 'range', min: 200, max: 300, step: 50 },
   },
 };
-
 export const Playground = ({ playgroundWidth, onChange, ...args }: StoryProps) => {
   // Parent component is in charge of the state management
-  const [value, setValue] = useState<string>(lotOptions[0]);
+  const [value, setValue] = useState<PaginationOptionValue>(lotOptions[0].value ?? lotOptions[0]);
```
sgrund14, line 43: is this `??` really necessary? isn't this just the hardcoded string from above, which you know is defined?

src/components/Pagination/Pagination.tsx
```diff
@@ -7,6 +7,20 @@ import Select from '../Select/Select';
 import ChevronRight from '../../assets/chevronRight.svg?react';
 import IconButton from '../IconButton/IconButton';
 import { ButtonVariants } from '../Button/types';
+import { determineOptionValue, findOptionFromSelectString } from './utils';
+
+export type PaginationOptionValue = string | number;
+
+export interface PaginationOption {
```
sgrund14, line 14: dup of changes in https://github.com/PhillipsAuctionHouse/seldon/pull/340#pullrequestreview-2343217552

src/components/Pagination/utils.tsx
```diff
@@ -0,0 +1,11 @@
+import { PaginationOption, PaginationOptionValue } from './Pagination';
+
+export const determineOptionValue = (option: PaginationOption | PaginationOptionValue) => {
```
sgrund14, line 3: would we ever want to export this from seldon? I guess not, since `onChange` is just getting passed the value...

sgrund14 (approved):

src/components/Pagination/Pagination.stories.tsx
scottdickerson: I think I could have used `lotOptions[0].value!` as well typescript was griping at me here

PR #340, opened 2024-09-27
fix(breadcrumb): L3-3900 breadcrumb style fixes

scottdickerson:
**Jira ticket**

[Jira Ticket ID](https://phillipsauctions.atlassian.net/browse/L3-3900)

**Screenshots**
| Before | After |
| ------ | ----- |
| Truncates Too Early | Uses full available width |
| image | image |

**Summary**

Changes the prop model for `Breadcrumb` to  remove the capability to truncate at a character length as this doesn't use the full space available for the Breadcrumb at responsive screen sizes before truncating.  It instead uses css ellipsis overflow to do this so if there is room to show the full text we do.  In addition truncating at the character count is bad for accessibility because screen readers will only see the truncated text and not the full text. A few more style fixes are included

This [PR](https://github.com/PhillipsAuctionHouse/seldon/pull/339) needs to be merged first so we only see these changes
- change(Breadcrumb/BreadcrumbItem): remove `truncateLength` prop
- change(_breadcrumb.scss): Add focus outline.  Make sure the truncation works on the truncate index with pure css overflow
- change(_iconButton): remove redundant styles, and fix `secondary` variant to be the right size

- Verify the Breadcrumb story truncates the `truncateIndex` correctly and leaves the other items at full length
- Check the focus outline and hover state of the `secondary` IconButton variant.

**Regression Test**

**Evidence of testing**

- With truncation
image

New Components

### DIFF
--- a/src/components/Pagination/Pagination.tsx
+++ b/src/components/Pagination/Pagination.tsx
@@ -7,6 +7,20 @@ import Select from '../Select/Select';
 import ChevronRight from '../../assets/chevronRight.svg?react';
 import IconButton from '../IconButton/IconButton';
 import { ButtonVariants } from '../Button/types';
+import { determineOptionValue, findOptionFromSelectString } from './utils';
+
+export type PaginationOptionValue = string | number;
+
+export interface PaginationOption {
+  /**
+   * label is translatable
+   */
+  label: string;
+  /*
+   * Each Pagination Option value must be unique within the Pagination
+   */
+  value: string | number;
+}
 
 export interface PaginationProps extends Omit<ComponentProps<'div'>, 'onChange'> {
   /**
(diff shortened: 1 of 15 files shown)


src/components/Pagination/Pagination.tsx
```diff
@@ -20,19 +34,19 @@ export interface PaginationProps extends Omit<ComponentProps<'div'>, 'onChange'>
   /**
    * Options Array of string
    */
-  options?: string[];
+  options: (PaginationOption | PaginationOptionValue)[];
```
sgrund14, line 37: non-blocking but should we just force this to be `PaginationOptionValue` instead of the mixing and matching?

sgrund14 (approved):

aahill50 (approved):
LGTM

SRangineni88 (approved):
LGTM..

PR #345, opened 2024-09-30
feat(subscribe): L3-3829 add ability to autofocus the input

scottdickerson:
**Jira ticket**

[Jira Ticket ID](https://phillipsauctions.atlassian.net/browse/L3-3829)

**Screenshots**
| Before | After |
| ------ | ----- |
| image | image |

**Summary**

Users would expect to be focused by default on the email input when the subscribe form opens, this change adds a property to be set to allow it.

- change(*.scss): fix input component focus states to match others (outline black)
- change(Subscribe): add autoFocus prop that focuses on the Input when set

- Popup the Drawer and Subscribe stories and verify that the text input field is focused
- Verify when you close the Drawer, whichever element on the page had focus previously now has had it reset to them.

**Regression Test**

- Interact with the Input, Select, Dropdown components still look right when interacting with them with both mouse and keyboard.

**Evidence of testing**

image

New Components

### DIFF
--- a/src/patterns/Subscribe/Subscribe.tsx
+++ b/src/patterns/Subscribe/Subscribe.tsx
@@ -72,6 +72,7 @@ export interface SubscribeProps extends React.HTMLAttributes<HTMLFormElement> {
  */
 
 const Subscribe = ({
+  autoFocus = true,
   blurb,
   buttonText = 'Sign Up',
   buttonProps,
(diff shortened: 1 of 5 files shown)


constantinehuzenko (approved):
LGTM

src/patterns/Subscribe/Subscribe.tsx
```diff
@@ -72,6 +72,7 @@ export interface SubscribeProps extends React.HTMLAttributes<HTMLFormElement> {
  */
 
 const Subscribe = ({
+  autoFocus = true,
```
sgrund14, line 75: would this ever get consumed somewhere like the footer?

might be weird to autofocus an element that's off screen unintentionally, if this is defaulting to true

scottdickerson: We're only planning to show it in some kind of Drawer at this point so I think this is a good default.

PR #347, opened 2024-09-30
feat(header): L3-3688 Highlight User Input In Search Results

adietrich3074:
**Jira ticket**

[L3-3688](https://phillipsauctions.atlassian.net/browse/L3-3688)

**Screenshots**
Screenshot 2024-09-30 161613

**Summary**

Added user input highlight and fixed search expansion animation on mobile

New Components

[L3-3688]: https://phillipsauctions.atlassian.net/browse/L3-3688?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Search/SearchResults/SearchResults.tsx
+++ b/src/components/Search/SearchResults/SearchResults.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import { px } from '../../../utils';
 import Link from '../../Link/Link';
 import { LinkVariants } from '../../Link';
+import { HTMLParser } from '../../HTMLParser';
 
 export interface SearchResult {
   id: string;
@@ -15,6 +16,7 @@ export interface SearchResultsProps extends React.HTMLAttributes<HTMLElement> {
   children?: React.ReactNode;
   loadingText?: string;
   onKeyDown?: (event: React.KeyboardEvent<HTMLAnchorElement>) => void;
+  userInputValue?: string;
 }
 
 const SearchResults = ({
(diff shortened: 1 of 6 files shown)


src/components/Search/SearchResults/SearchResults.tsx
```diff
@@ -23,9 +25,18 @@ const SearchResults = ({
   children,
   loadingText = 'Loading...',
   onKeyDown,
+  userInputValue = '',
 }: React.PropsWithChildren<SearchResultsProps>) => {
   const hasResults = Array.isArray(autoCompleteResults) && autoCompleteResults.length > 0;
 
+  function formatSearchLabel(label: string, searchQuery: string): React.ReactNode {
```
scottdickerson, line 32: this is just a personal preference thing, but I like `const formatSearchLabel = (label:string, searchQuery: string) => better`.  Also you shouldn't have to type the return type as `React.ReactNode` I think it will be inferred from the return statement below.

src/components/Search/SearchResults/SearchResults.tsx
```diff
@@ -23,9 +25,18 @@ const SearchResults = ({
...
+    return (
+      <HTMLParser
+        html={`<span class='${px}-search__result__label'>${label.replace(new RegExp(searchQuery, 'gi'), (match) => `<strong>${match}</strong>`)}</span>`}
```
scottdickerson, line 35: I love that you chose to use the `HTMLParser` here.

scottdickerson:
for some reason on mobile, the View all link font size no longer matches the font of the search results: 
image
Was that intentional?

scottdickerson (approved):
LGTM

PR #349, opened 2024-10-01
fix(searchresult): L3-3688 highlight user input in search result breaking Remix with lib is not defined

scottdickerson:
**Jira ticket**

[Jira Ticket ID](https://phillipsauctions.atlassian.net/browse/L3-3688)

**Summary**

SearchResult was using the `<HTMLParser` component in seldon which unfortunately breaks the SSR in Remix with the good old `lib is not defined` error.

- remove(HTMLParser): remove the local component since it doesn't work server-side
- change(SearchResults): handle the React element wrapping without using HTMLParser

- Test SearchResults in the story, searching on "Yet"

**Regression Test**

**Evidence of testing**

image

New Components


adietrich3074 (approved):
👍

PR #355, opened 2024-10-03
feat(detail): L3-4057 detail list improvement

constantinehuzenko:
**Jira ticket**

[L3-4057](https://phillipsauctions.atlassian.net/browse/L3-4057)

**Screenshots**
| Before | After |
| ------ | ----- |
| N/A | N/A |

**Summary**

Updated `DetailList` and `Detail`

- src/components/Detail/Detail.tsx: updated type of `value` and `label` to take not only `sting` but `ReactNode` as a child, added optional and default props
- src/patterns/DetailList/DetailList.tsx: added single element support, removed `undefined` class name, removed generic `index` key, added optional and default props
- src/patterns/DetailList/index.ts: exported type enum

- verify that you can add one element to `DetailList`
- verify that `Detail` class name is not undefined in DOM
- verify that you can pass `ReactNode` to `Detail`

**Regression Test**

**Evidence of testing**

- Post logs, screenshots, etc

New Components

[L3-4057]: https://phillipsauctions.atlassian.net/browse/L3-4057?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Detail/Detail.tsx
+++ b/src/components/Detail/Detail.tsx
@@ -1,15 +1,16 @@
-import { ComponentProps, forwardRef } from 'react';
+import React, { ComponentProps, forwardRef } from 'react';
 import { getCommonProps } from '../../utils';
+import classNames from 'classnames';
 
 export interface DetailProps extends ComponentProps<'div'> {
   /*
    * Label that appears on the left side of the Detail component
    */
-  label: string;
+  label?: React.ReactNode;
   /*
    * Value that appears on the right side of the Detail component
    */
-  value: string;
+  value?: React.ReactNode;
 }
 /**
  * ## Overview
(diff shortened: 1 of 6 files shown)


src/patterns/DetailList/DetailList.tsx
```diff
@@ -44,18 +44,22 @@ const DetailList = forwardRef<HTMLDListElement, DetailListProps>(
...
-              [`${px}-columns`]: alignment === DetailListAlignment.columns,
-              [`${px}-justified`]: alignment === DetailListAlignment.justified,
-            })}
-            key={getDetailKey(child, index)}
-          >
-            {child}
-          </div>
-        ))}
+        {childrenArray?.map((child) => {
+          const key = isValidElement(child) ? child?.key : null;
```
sgrund14, line 48: is it guaranteed that child will have a key?

src/patterns/DetailList/DetailList.tsx
```diff
@@ -31,8 +30,9 @@ export interface DetailListProps extends ComponentProps<'dl'> {
  *
  */
 const DetailList = forwardRef<HTMLDListElement, DetailListProps>(
-  ({ alignment, className, children, hasSeparators, ...props }, ref) => {
+  ({ alignment = DetailListAlignment.columns, className, children, hasSeparators = false, ...props }, ref) => {
```
scottdickerson, line 33: I think making the other Alignment the default is better as that seems to be more common in our current Figmas

src/patterns/DetailList/DetailList.tsx
scottdickerson: I actually think the `key={getDetailKey(child, index)}` is perfect code the right way to do it

scottdickerson (changes requested):
just a few small changes

src/components/Detail/Detail.tsx
```diff
@@ -1,15 +1,16 @@
...
+import React, { ComponentProps, forwardRef } from 'react';
 import { getCommonProps } from '../../utils';
+import classNames from 'classnames';
 
 export interface DetailProps extends ComponentProps<'div'> {
   /*
    * Label that appears on the left side of the Detail component
    */
-  label: string;
+  label?: React.ReactNode;
```
sgrund14, line 9: if we are using `label` and `value` to derive the key, they should be required

otherwise, should make `key` required

scottdickerson: yeah I would argue detail without a label and value doesn't make sense as well

scottdickerson (approved):
LGTM once @sgrund14 request is addressed

src/components/Detail/Detail.tsx
sgrund14: are there any cases where you'd have one but not the other?

sgrund14 (approved):

PR #360, opened 2024-10-07
feat(sale header banner): L3-2394 create sale header banner

smarks8:
**Jira ticket**

[L3-2394](https://phillipsauctions.atlassian.net/browse/L3-2394)

**Screenshots**
Screenshot 2024-10-07 090653
Screenshot 2024-10-07 090714
Screenshot 2024-10-07 090839
Screenshot 2024-10-07 090900

**Summary**

Create sale header banner component for sale page

- Creation of new component with all associated files: tests, storybook, types, styles

- Playground in storybook

**Regression Test**

**Evidence of testing**

Screenshot 2024-10-07 091150

New Components

[L3-2394]: https://phillipsauctions.atlassian.net/browse/L3-2394?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


davidicus:
@smarks8 The button seems a little tight to the text. It looks like we need padding here.

src/patterns/SaleHeaderBanner/SaleHeaderBanner.tsx
```diff
@@ -0,0 +1,106 @@
...
+                <Text variant={TextVariants.heading5}>17 Hours</Text>
+              </div>
+            )}
+            <Text variant={TextVariants.title1}>{auctionTitle}</Text>
+            <Text variant={TextVariants.string2} className={`${baseClassName}__location`}>
+              {location}
+            </Text>
+            <div
+              className={
+                isClosed ? `${baseClassName}__occurance-details-closed` : `${baseClassName}__occurance-details`
```
davidicus, line 77: This should probably use the classnames package for adding classes based off a condition. I would probably just have base class name render all the time and then the modifier if the `isClosed` condition is true. 

Following the BEM convention the modifier would be `${baseClassName}__occurance-details--closed` with a double dash. 
```suggestion
                classnames(`${baseClassName}__occurance-details`, {`${baseClassName}__occurance-details--closed`: isClosed})
```

src/patterns/SaleHeaderBanner/SaleHeaderBanner.tsx
```diff
@@ -0,0 +1,106 @@
...
+        ></img>
+        <PageMargin>
+          <div className={`${baseClassName}__stack`}>
+            {isOpenForBidding && (
+              <div id="PLACEHOLDER FOR TIMER COMPONENT" className={`${baseClassName}__countdown-container`}>
+                <Text variant={TextVariants.heading5}>Lots Close in</Text>
+                <Text variant={TextVariants.heading5}>2 Days</Text>
+                <Text variant={TextVariants.heading5}>17 Hours</Text>
+              </div>
+            )}
```
davidicus, line 70: I wonder if we should make this composable? These string will need a way to be translated. We either need to include as props for this component or allow for composability of a timer that will include translated strings.

src/patterns/SaleHeaderBanner/SaleHeaderBanner.tsx
```diff
@@ -0,0 +1,106 @@
...
+    const isClosed = auctionState === AuctionState.past;
+
+    return (
+      <header {...commonProps} className={classnames(baseClassName, className)} {...props} ref={ref}>
+        <img
+          id="PLACEHOLDER FOR IMAGE COMPONENT"
+          alt={auctionTitle}
+          className={`${baseClassName}__image`}
+          src={imageSrcUrl}
+        ></img>
```
davidicus, line 61: closing image tag is not needed.

src/patterns/SaleHeaderBanner/_saleHeaderBanner.scss
```diff
@@ -0,0 +1,69 @@
...
+
+  &__location {
+    font-variation-settings: 'wght' 600;
+    margin: $margin-xsm 0;
+  }
+
+  &__occurance-details,
+  &__occurance-details-closed {
+    display: flex;
+    margin-bottom: 24px;
```
davidicus, line 48: token?

src/patterns/SaleHeaderBanner/_saleHeaderBanner.scss
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  // remove when adding in actual countdown timer
+  &__countdown-container {
+    align-items: center;
+    border-bottom: 1px solid $light-gray;
+    display: flex;
+    gap: $spacing-sm;
+    justify-content: center;
+    margin-bottom: 24px;
```
davidicus, line 36: Does this have a token for this value in the figma?

src/patterns/SaleHeaderBanner/_saleHeaderBanner.scss
```diff
@@ -0,0 +1,69 @@
...
+
+  @media screen and (min-width: $breakpoint-md) {
+    flex-direction: row-reverse;
+  }
+
+  // update when adding image component
+  &__image {
+    aspect-ratio: 16 / 9;
+
+    @media screen and (min-width: $breakpoint-md) {
```
davidicus, line 15: FYI we do have a media query mixin.

src/patterns/SaleHeaderBanner/_saleHeaderBanner.scss
davidicus: Yeah, ok, looking at the figma  I see they do not have an associated token. Can we just change these to Rem units? Thanks!

src/patterns/SaleHeaderBanner/SaleHeaderBanner.tsx
```diff
@@ -0,0 +1,108 @@
...
+                <Text variant={TextVariants.string2} className={`${baseClassName}__date`}>
+                  {date}
+                </Text>
+              </div>
+              {isClosed && (
+                <>
+                  <hr />
+                  <div className={`${baseClassName}__occurance-details-text`}>
+                    <Text variant={TextVariants.string2}>Browse Upcoming Sale</Text>
+                    <Link href="/calendar">View Calendar</Link>
```
scottdickerson, line 93: same things for translations here and possibly a composable component.  I.e. SaleHeaderCountdown, SaleHeaderBrowseAuctions, etc

scottdickerson (changes requested):
hoping that making this more composable will make the prop model simpler

src/patterns/SaleHeaderBanner/SaleHeaderBanner.tsx
sgrund14: wondering if this actually makes sense as a seldon component, rather than living in remix as a composition of different seldon components (which are already composable)

or will this be used by phillips public?

src/patterns/SaleHeaderBanner/SaleHeaderBanner.tsx
```diff
@@ -0,0 +1,109 @@
...
+   */
+  location: string;
+  /**
+   * Depending on auction state, When does the auction open or close
+   */
+  date: string;
+  /**
+   * Clarifies the date based on the auction state
+   */
+  occuranceLabel: string;
```
scottdickerson, line 31: minor but this is typically spelled `occurrence`

src/patterns/SaleHeaderBanner/SaleHeaderBanner.tsx
```diff
@@ -0,0 +1,109 @@
...
+            >
+              <div className={`${baseClassName}__occurance-details-text`}>
+                <Text variant={TextVariants.string2}>{occuranceLabel}</Text>
+                <Text variant={TextVariants.string2} className={`${baseClassName}__date`}>
+                  {date}
+                </Text>
+              </div>
+              {isClosed && children}
+            </div>
+            {!isClosed && <Button className={`${baseClassName}__cta`}>{ctaLabel}</Button>}
```
scottdickerson, line 99: do we need a callback here so consumers can handle the register to bid button?

src/patterns/SaleHeaderBanner/SaleHeaderBanner.stories.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+export default meta;
+export const Playground = (props: SaleHeaderBannerProps) => <SaleHeaderBanner {...props} />;
+
+Playground.args = {
+  auctionTitle: 'Modern & Contemporary Art: Online Auction, New York',
+  date: '10:00am EDT, 14 Sep 2024',
+  occuranceLabel: 'Begins',
+  location: 'New York',
+  auctionState: AuctionState.preSale,
```
scottdickerson, line 21: can you set `argTypes` so that the `auctionState` is a dropdown option that's selectable, there should be a few examples using the `TextVariants` that you can copy

src/patterns/SaleHeaderBanner/SaleHeaderBanner.tsx
```diff
@@ -0,0 +1,120 @@
...
+import { Text, TextVariants } from '../../components/Text';
+import { PageContentWrapper as PageMargin } from '../../components/PageContentWrapper';
+import Button from '../../components/Button/Button';
+
+// You'll need to change the ComponentProps<"htmlelementname"> to match the top-level element of your component
+export interface SaleHeaderBannerProps extends ComponentProps<'div'> {
+  /**
+   * What is the title of the auction?
+   */
+  auctionTitle: string;
```
scottdickerson, line 15: can we make these `ReactNode` in case we need to pass custom elements (location, date, occuranceLabel and ctaLabel) probably as well

src/patterns/SaleHeaderBanner/SaleHeaderBanner.tsx
scottdickerson: I think there are some benefits to it being in Seldon, first of all, the Easel Design System from Figma references it in their Pattern section.  I think we need to examine some of our other components in Remix (LotNavigation, etc) for potentially moving to Seldon as well.

scottdickerson (approved):
LGTM

PR #361, opened 2024-10-07
feat(imagecomponent): L3-1466 seldon image component

sgrund14:
**Jira ticket**

[L3-1466](https://phillipsauctions.atlassian.net/browse/L3-1466)

**Screenshots**

Screenshot 2024-10-07 at 10 10 45 AM

**Summary**

Seldon image component
Support different aspect ratios & fill modes
With optional blur background

TBD loading & error states

- SeldonImage component, stories, and tests

**Regression Test**

**Evidence of testing**

- Post logs, screenshots, etc

New Components

[L3-1466]: https://phillipsauctions.atlassian.net/browse/L3-1466?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Carousel/CarouselDots.tsx
+++ b/src/components/Carousel/CarouselDots.tsx
@@ -47,8 +47,6 @@ const CarouselDots = forwardRef<HTMLDivElement, CarouselDotsProps>(
       [api, onSlideChange],
     );
 
-    console.log(api?.slideNodes());
-
     const onInit = useCallback((emblaApi: EmblaCarouselType) => {
       setScrollSnaps(emblaApi.scrollSnapList());
     }, []);
--- a/src/components/Carousel/CarouselDots.tsx
+++ b/src/components/Carousel/CarouselDots.tsx
@@ -47,6 +47,8 @@ const CarouselDots = forwardRef<HTMLDivElement, CarouselDotsProps>(
       [api, onSlideChange],
     );
 
+    console.log(api?.slideNodes());
+
     const onInit = useCallback((emblaApi: EmblaCarouselType) => {
       setScrollSnaps(emblaApi.scrollSnapList());
     }, []);
@@ -103,13 +105,13 @@ const CarouselDots = forwardRef<HTMLDivElement, CarouselDotsProps>(
       >
         <div className={`${baseClassName}-pagination-container`}>
           <div className={`${baseClassName}-pagination-container-inner`}>
-            {visibleDots.map((snapPoint, index) => {
+            {visibleDots.map((_, index) => {
               const actualIndex = scrollSnaps.indexOf(visibleDots[index]);
               const isSelected = selectedIndex === actualIndex;
 
               return (
                 <button
-                  key={`${componentId}-dot-${snapPoint}`}
+                  key={`${componentId}-dot-${index}`}
                   role="button"
                   onClick={() => onDotButtonClick(actualIndex)}
                   className={classNames(`${baseClassName}-pagination-dot-container`)}
--- a/src/components/SeldonImage/SeldonImage.tsx
+++ b/src/components/SeldonImage/SeldonImage.tsx
@@ -1,27 +1,48 @@
-import { ComponentProps, forwardRef, useImperativeHandle, useRef } from 'react';
+import { ComponentProps, forwardRef, useState } from 'react';
 import { getCommonProps } from '../../utils';
 import classnames from 'classnames';
 
 type AspectRatio = '16/9' | '1/1' | 'none';
 
 export interface SeldonImageProps extends ComponentProps<'div'> {
+  /**
+   * The aspect ratio of the image container.
+   */
   aspectRatio?: AspectRatio;
+  /**
+   * The resize behavior of the image (see CSS object-fit).
+   */
   objectFit?: 'contain' | 'cover' | 'fill' | 'none' | 'scale-down';
-  useBlurBackground?: boolean;
-  imageClassName?: string;
-  imageStyle?: React.CSSProperties;
+  /**
+   * Whether the image has a blur background covering the background of the image container
+   */
+  hasBlurBackground?: boolean;
+  /**
+   * The image to display.
+   */
   src: string;
+  /**
+   * The alt text of the image.
+   */
   alt?: string;
+  /**
+   * The class name of the child img element.
+   */
+  imageClassName?: string;
+  /**
+   * The style of the child img element.
+   */
+  imageStyle?: React.CSSProperties;
 }
 
 /**
  * ## Overview
  *
- * Overview of this widget
+ * Component for displaying an image with optional blur background, aspect ratio, and object fit.
  *
- * [Figma Link](Add Figma URL here)
+ * [Figma Link](https://www.figma.com/design/hMu9IWH5N3KamJy8tLFdyV/EASEL-Compendium%3A-Tokens%2C-Components-%26-Patterns?m=auto&node-id=4501-64590)
  *
- * [Storybook Link](Point back to yourself here)
+ * [Storybook Link](https://phillips-seldon.netlify.app/?path=/docs/components-seldonimage--overview)
  */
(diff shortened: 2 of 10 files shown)


src/components/SeldonImage/SeldonImage.tsx
```diff
@@ -0,0 +1,90 @@
...
+        )}
+        <img
+          className={classnames(`${baseClassName}-img`, imgObjectFitClass, imgHiddenClass, imageClassName)}
+          style={imageStyle}
+          ref={imgRef}
+          src={src}
+          onLoad={() => {
+            imgRef.current?.classList.remove(imgHiddenClass);
+            containerRef.current?.classList.remove(containerHiddenClass);
+            blurRef.current?.classList.remove(blurHiddenClass);
```
davidicus, line 83: Should this all just be done with state and then these classes updated with classnames package

src/components/SeldonImage/SeldonImage.tsx
```diff
@@ -0,0 +1,90 @@
...
+      return document.createElement('div');
+    });
+
+    const containerAspectRatioClass =
+      aspectRatio !== 'none' ? `${baseClassName}--aspect-ratio-${aspectRatio.replace('/', '-')}` : '';
+    const containerHiddenClass = imgRef.current?.complete ? '' : `${baseClassName}--hidden`;
+    const imgObjectFitClass = objectFit !== 'none' ? `${baseClassName}-img--object-fit-${objectFit}` : '';
+    const imgHiddenClass = imgRef.current?.complete ? '' : `${baseClassName}-img--hidden`;
+    const blurHiddenClass = imgRef.current?.complete ? '' : `${baseClassName}-blur--hidden`;
+
```
davidicus, line 58: Should this all be done with classnames package?

davidicus (commented):
Looks great. Some minor comments.

src/components/SeldonImage/SeldonImage.tsx
sgrund14: yea idk why I went for refs with this one, something about not using react state, but it's probably more legible that way

src/components/Carousel/CarouselDots.tsx
```diff
@@ -103,13 +103,13 @@ const CarouselDots = forwardRef<HTMLDivElement, CarouselDotsProps>(
...
           <div className={`${baseClassName}-pagination-container-inner`}>
-            {visibleDots.map((snapPoint, index) => {
+            {visibleDots.map((_, index) => {
               const actualIndex = scrollSnaps.indexOf(visibleDots[index]);
               const isSelected = selectedIndex === actualIndex;
 
               return (
                 <button
-                  key={`${componentId}-dot-${snapPoint}`}
+                  key={`${componentId}-dot-${index}`}
```
sgrund14, line 112: added this back for now, otherwise it's throwing a key warning
`snapPoint` array gets initialized with all zero's, so it has identical keys on first render

sgrund14: i can make it required, or default it to the empty string

was trying to make it match the img api

src/components/SeldonImage/SeldonImage.tsx
```diff
@@ -0,0 +1,115 @@
...
+        )}
+        <img
+          className={classnames(`${baseClassName}-img`, imageClassName, {
+            [`${baseClassName}-img--hidden`]: loadingState === 'loading',
+            [`${baseClassName}-img--object-fit-${objectFit}`]: objectFit !== 'none',
+          })}
+          style={imageStyle}
+          src={src}
+          alt={alt}
+          role="img"
```
davidicus, line 99: This may be better set on the containing div. I believe this [role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/img_role) is used for container of multiple elements that make up an image.

davidicus (approved):

davidicus:
Looks great!

src/components/SeldonImage/SeldonImage.tsx
```diff
@@ -0,0 +1,117 @@
...
+    const { className: baseClassName, ...commonProps } = getCommonProps(props, 'SeldonImage');
+    const imgRef = useRef<HTMLImageElement>(null);
+
+    const [loadingState, setLoadingState] = useState<'loading' | 'loaded' | 'error'>('loading');
+
+    useEffect(() => {
+      if (imgRef.current?.complete) {
+        setLoadingState('loaded');
+      }
+    }, []);
```
scottdickerson, line 71: Why did we need this in addition to the `onLoad` below? I suspect this won't actually get fired since the `imgRef.current` is never set on initial render.

sgrund14: we may not, I think I threw this in there to fix this weird flash I was seeing in browser but it may have just been a browser cache thing, and it was making it seem like `onLoad` was firing inconsistently

it actually is defined though, since the first `useEffect` fires after first paint

src/components/SeldonImage/SeldonImage.tsx
```diff
@@ -0,0 +1,117 @@
...
+            className={classnames(`${baseClassName}-blur`, {
+              [`${baseClassName}-blur--hidden`]: loadingState === 'loading',
+            })}
+            style={{ backgroundImage: `url(${src})` }}
+          />
+        )}
+        <img
+          className={classnames(`${baseClassName}-img`, imageClassName, {
+            [`${baseClassName}-img--hidden`]: loadingState === 'loading',
+            [`${baseClassName}-img--object-fit-${objectFit}`]: objectFit !== 'none',
```
scottdickerson, line 96:
```suggestion
            [`${baseClassName}-img--hidden`]: loadingState === 'loading' || loadingState === 'error',
```

src/components/SeldonImage/SeldonImage.tsx
```diff
@@ -0,0 +1,117 @@
...
+      >
+        {hasBlurBackground && (
+          <div
+            className={classnames(`${baseClassName}-blur`, {
+              [`${baseClassName}-blur--hidden`]: loadingState === 'loading',
+            })}
+            style={{ backgroundImage: `url(${src})` }}
+          />
+        )}
+        <img
```
scottdickerson, line 93:
```suggestion
        {loadingState === 'error' && <div className={`${baseClassName}--error`}>{`${errorText}: ${src}`}</div>}
```

src/components/SeldonImage/SeldonImage.tsx
```diff
@@ -0,0 +1,117 @@
...
+  alt: string;
+  /**
+   * The class name of the child img element.
+   */
+  imageClassName?: string;
+  /**
+   * The style of the child img element.
+   */
+  imageStyle?: React.CSSProperties;
+}
```
scottdickerson, line 36:
```suggestion
  errorText?: string;
```

src/components/SeldonImage/SeldonImage.tsx
```diff
@@ -0,0 +1,117 @@
...
+    {
+      className,
+      aspectRatio = 'none',
+      objectFit = 'none',
+      hasBlurBackground = false,
+      imageClassName,
+      imageStyle,
+      src,
+      alt,
+      ...props
```
scottdickerson, line 58:
```suggestion
      errorText = 'Error loading image',
```

src/components/SeldonImage/_seldonImage.scss
```diff
@@ -0,0 +1,61 @@
...
+
+  &--aspect-ratio-1-1 {
+    aspect-ratio: 1 / 1;
+  }
+
+  &--hidden {
+    background-color: #eee;
+  }
+
+  &-img {
```
scottdickerson, line 19:
```suggestion
  &--error {
    text-align: center;
  }
```

scottdickerson (approved):
LGTM

PR #363, opened 2024-10-07
feat(Tabs): L3-2396 Create Tabs component

SRangineni88:
**Jira ticket**

[L3-2396](https://phillipsauctions.atlassian.net/browse/L3-2396)

**Screenshots**
|image

**Summary**

Tabs component for “Overview” and “Browse”
Component can take any number of tabs.
Support any tab to be selected on page render
The Full Bleed on the underline

**Regression Test**

**Evidence of testing**

- Post logs, screenshots, etc

New Components

[L3-2396]: https://phillipsauctions.atlassian.net/browse/L3-2396?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sgrund14 (commented):
could in theory export this as four separate components too, like how it's done here: https://github.com/shadcn-ui/ui/blob/main/apps/www/registry/default/ui/tabs.tsx

`Tabs`, `TabsList`, `TabsTrigger`, and `TabsContent`

Then the stories could demonstrate how to compose them as you do in the current main `Tabs` component

Similarly to the [Collapsible](https://github.com/PhillipsAuctionHouse/seldon/blob/main/src/components/Collapsible/index.ts) component

Not necessarily a blocker though

Thoughts @scottdickerson @davidicus ?

src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+const TabsComponent = ({ className, tabs = [], defaultValue, ...props }: TabsComponentProps) => {
+  const { className: baseClassName, ...commonProps } = getCommonProps(props, 'TabsComponent');
+  return (
+    <Tabs.Root
+      defaultValue={defaultValue || tabs[0].value}
+      {...commonProps}
+      className={classnames(`${baseClassName}`, className)}
+    >
+      <Tabs.List aria-label="Tabs" className={`${baseClassName}__tabs-list`}>
```
davidicus, line 54: aria-label should be a prop that we can pass in a translated string

davidicus:
> could in theory export this as four separate components too, like how it's done here: https://github.com/shadcn-ui/ui/blob/main/apps/www/registry/default/ui/tabs.tsx
> 
> `Tabs`, `TabsList`, `TabsTrigger`, and `TabsContent`
> 
> Then the stories could demonstrate how to compose them as you do in the current main `Tabs` component
> 
> Similarly to the [Collapsible](https://github.com/PhillipsAuctionHouse/seldon/blob/main/src/components/Collapsible/index.ts) component
> 
> Not necessarily a blocker though
> 
> Thoughts @scottdickerson @davidicus ?

@sgrund14 I was thinking of multiple exported components that we can compose as well. I was wondering if it makes sense to simplify by having a `Tabs` container and then the children would be wrapped in the  `TabContent` component.

```js
<Tabs tabLabels={[{label: "Overview", value: "tab1"}, {label: "Browse lots", value: "tab2"}]} defaultValue="tab1">
  <TabContent value="tab1">
      {// content}
  </TabContent>
  <TabContent value="tab2">
      {// content}
  </TabContent>
</Tabs>

File: src/components/Tabs/Tabs.tsx [hunk 2]
```diff
@@ -0,0 +1,71 @@
...
+      {tabs.map((tab) => (
+        <Tabs.Content key={tab.value} value={tab.value} className={`${baseClassName}__tabs-content`}>
+          {tab.content}
+        </Tabs.Content>
+      ))}
+    </Tabs.Root>
+  );
+};
+
+export default TabsComponent;
```
davidicus, line 71: If we are exporting as Tabs I would like to name the component the same. I see you exported the radix component as Tabs which makes senes because you have to type it more but I like the explicit labeling of the external component as `TabsPrimitive` or `RadixTabs`. TabsPrimitive would match the naming convention started by Sam [here](https://github.com/PhillipsAuctionHouse/seldon/blob/main/src/components/Collapsible/Collapsible.tsx)

src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,71 @@
...
+ * @param {TabsComponentProps} props - The component props containing an array of tabs.
+ * @returns {JSX.Element} The rendered tab component.
+ */
+export interface TabsComponentProps extends ComponentProps<'div'> {
+  tabs: Tab[];
+  /**
+   * Specify the default tab
+   */
+  defaultValue?: string;
+}
```
davidicus, line 34: We may want to have an optional onClick handler that we call with the tab value for cases like updating the URL with the open tab

src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const { className: baseClassName, ...commonProps } = getCommonProps(props, 'TabsComponent');
+  return (
+    <Tabs.Root
+      defaultValue={defaultValue || tabs[0].value}
+      {...commonProps}
+      className={classnames(`${baseClassName}`, className)}
+    >
+      <Tabs.List aria-label="Tabs" className={`${baseClassName}__tabs-list`}>
+        {tabs.map((tab) => (
+          <Tabs.Trigger key={tab.value} value={tab.value} className={`${baseClassName}__tabs-trigger`}>
```
scottdickerson, line 56: I was hoping these `<Tabs.Trigger` would be their own component and consumers would be able to pass props to them directly as composable elements.  I.e. a `<Tabs`, `<Tab`, `<TabContent` component would be used by consumers

src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,71 @@
...
+            {tab.label}
```
scottdickerson, line 57: Should we use the `<Text` component here?

src/components/Tabs/_tabs.scss
```diff
@@ -0,0 +1,53 @@
...
+      color: $pure-black;
+    }
+  }
+
+  &__tabs-trigger[data-state='active'] {
+    border-bottom: 2px solid $pure-black;
+    color: $pure-black;
+    font-variation-settings: 'wght' 600;
+
+    &:hover {
```
scottdickerson, line 34: can we get a better focus state for keyboard navigation of these Tabs?  Currently it's the default blue outline which looks pretty bad.  I believe  something like this from breadcrumb 
```  &:focus-visible {
        border-radius: 5px;
        outline: 1px solid $medium-gray;
      }
```

scottdickerson (changes requested):
Requested some changes to the Tab component composition

src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,71 @@
...
+ * ## Overview
+ *
+ * Overview of Tabs component
+ *
+ * [Figma Link](https://www.figma.com/file/hMu9IWH5N3KamJy8tLFdyV?type=design&node-id=10838%3A22184&mode=dev)
+ *
+ * [Storybook Link](https://phillips-seldon.netlify.app/?path=/docs/components-tabs--overview)
+ */
+
+const TabsComponent = ({ className, tabs = [], defaultValue, ...props }: TabsComponentProps) => {
```
scottdickerson, line 46: could you wrap this component in `forwardRef` to this component as well as any of the composable components we come up with

scottdickerson: you'll typically get this for free if you use the create-component script:
https://github.com/PhillipsAuctionHouse/seldon/blob/2a1de144ec0be4a7d34c4d0745847030ad7cad00/.template/Component/ComponentName.tsx#L16

src/components/Tabs/TabsContainer.tsx
```diff
@@ -0,0 +1,87 @@
...
+ * Overview of Tabs component
+ *
+ * [Figma Link](https://www.figma.com/file/hMu9IWH5N3KamJy8tLFdyV?type=design&node-id=10838%3A22184&mode=dev)
+ *
+ * [Storybook Link](https://phillips-seldon.netlify.app/?path=/docs/components-tabs--overview)
+ */
+
+const TabsContainer = forwardRef<HTMLDivElement, TabsComponentProps>(
+  ({ className, tabs = [], tabListLabel = 'Sale Page Tabs', children, defaultValue, onTabClick, ...props }, ref) => {
+    const { className: baseClassName, ...commonProps } = getCommonProps(props, 'TabsComponent');
```
scottdickerson, line 58: could we change this to tabs container?

src/components/Tabs/index.ts
```diff
@@ -0,0 +1,3 @@
+export { default as TabsContainer, type TabsComponentProps } from './TabsContainer';
+export { default as TabsContent, type TabContentProps } from './TabsContent';
+export { default as TabTrigger, type TabTriggerProps } from './TabTrigger';
```
scottdickerson, line 3: we probably don't need to export this as it's only used internally

scottdickerson:
I'm seeing the tab label raise up by a few pixels when it's selected.  Maybe it needs a translucent bottom border? 


https://github.com/user-attachments/assets/b1dfe315-defb-4e7e-982b-8d27fd31d6fd

scottdickerson:
The focus state looks strange 
image
And should the Tab Content area be keyboard focusable at all?  
image

scottdickerson (approved):
LGTM

PR #376, opened 2024-10-15
fix(header): L3-4097 - fix padding and transition bug

davidicus:
**Jira ticket**

[L3-4097](https://phillipsauctions.atlassian.net/browse/L3-4097)

**Screenshots**
| Before | After |
| ------ | ----- |
| image | image |

**Summary**

This PR addresses a few discrepancies between implementation and Figma design. It also addresses a transition bug on resize.

- Added media query to only show underlines for header links on desktop in `src/components/Link/_link.scss`
- Corrected token value in `src/components/Navigation/NavigationList/_navigationList.scss` to match figma
- Removed border for bottom link accordion and added the one for language accordion item. I also swapped out units for transition bug in `src/site-furniture/Header/_header.scss`
- change(accordionitem): add special class for locked content with padding and make the "Large" variant have a larger click target on the trigger
- change(_accordion.scss): move the bottom margin from the overall accordion item, to the individual "label" and "content" so we can adjust for the different open/closed state and the locked vs unlocked content
- change(Header): use Large AccordionVariant
- change(navigationList): remove the item padding as it's no longer needed
- change(languageSelector): remove the nav__list padding and use the accordion padding. use Large AccordionVariant
- change(footer): remove the item padding and ul padding as it's no longer needed. Use the OOTB label padding. use Large AccordionVariant

- Make sure header story matches the [figma](https://www.figma.com/design/hMu9IWH5N3KamJy8tLFdyV/EASEL-Compendium%3A-Tokens%2C-Components-%26-Patterns?node-id=10572-6185&node-type=canvas&m=dev)

**Regression Test**

- Rest of header, like the desktop view, should all look normal
- Mobile Header and Mobile Footer and Mobile Language Selector should look correct

New Components

[L3-4097]: https://phillipsauctions.atlassian.net/browse/L3-4097?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,17 @@
+## [1.77.2](https://github.com/PhillipsAuctionHouse/seldon/compare/v1.77.1...v1.77.2) (2024-10-15)
+
+
+### Bug Fixes
+
+* **pagination:** L3-4039 improve async page navigation ([#370](https://github.com/PhillipsAuctionHouse/seldon/issues/370)) ([8f1a497](https://github.com/PhillipsAuctionHouse/seldon/commit/8f1a497fa972356f2fc1048d8263f69ac221b624))
+
+## [1.77.1](https://github.com/PhillipsAuctionHouse/seldon/compare/v1.77.0...v1.77.1) (2024-10-15)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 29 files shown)


adietrich3074 (approved):
Looks great! Thanks for the help on the header 😄

scottdickerson:
one thing I'm noticing is that the full Nav Trigger for Calendar is a clickable target, but only a small part of the Auctions or Department is clickable.  It seems like the whole area for the accordions (in addition to padding) should be clickable

scottdickerson (changes requested):
Accordion clickability is confusing at these larger sizes

scottdickerson (approved):
LGTM

PR #381, opened 2024-10-17
feat(countdown): L3-3809 add new countdown component

smarks8:
**Jira ticket**

[L3-3809](https://phillipsauctions.atlassian.net/browse/L3-3809)

**Screenshots**
Screenshot 2024-10-17 115058
image

**Summary**

Add new countdown component for sale header banner

- Create new component and all related files
- Add new component to sale header banner, remove previous placeholder

- `Countdown` component in storybook with playground args
- See `SaleHeaderBanner` component as well, swapped out placeholder  

**Regression Test**

**Evidence of testing**

image

New Components

[L3-3809]: https://phillipsauctions.atlassian.net/browse/L3-3809?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -7,6 +7,9 @@ export * from './assets/icons';
 // 📑 Pages
 export { default as Page } from './pages/Page';
 
+// utils
+export * from './utils/hooks';
+
 // ⚛️ Components
 export { default as Button, type ButtonProps } from './components/Button/Button';
 export { ButtonVariants } from './components/Button/types';
@@ -79,7 +82,5 @@ export * from './components/PinchZoom';
 export * from './components/Tabs';
 export * from './components/SeldonImage';
 export * from './patterns/SaleHeaderBanner';
-
-// utils
-export * from './utils/hooks';
 export * from './components/Countdown';
+export * from './components/Countdown/types';


sgrund14:
> wondering if we should use the custom code or utilize a library like date-fns
 I think the custom code is fine 🤷 

do we want a fixed width on the numbers so as to avoid little shifts when the numbers change?

https://github.com/user-attachments/assets/9d8eb799-f96e-460b-86ca-ca3aa0a2f80b

src/components/Countdown/Countdown.tsx
```diff
@@ -0,0 +1,151 @@
...
+          {timeLeft.days <= 0 && (
+            <span className={`${baseClassName}__time-item`}>
+              <span className={`${baseClassName}__time-item--numbers`}>
+                {timeLeft.minutes.toString().padStart(2, '0')}
+              </span>{' '}
+              <span>{minutesLabel}</span>
+            </span>
+          )}
+          {timeLeft.hours <= 0 && timeLeft.days === 0 && (
+            <span className={`${baseClassName}__time-item`}>
```
sgrund14, line 135: spacing a little tight between numbers & labels?

Screenshot 2024-10-17 at 5 40 17 PM

src/components/Countdown/Countdown.tsx
```diff
@@ -0,0 +1,151 @@
+import { ComponentProps, forwardRef, useCallback, useEffect, useState } from 'react';
+import { getCommonProps } from '../../utils';
+import classnames from 'classnames';
+
+// You'll need to change the ComponentProps<"htmlelementname"> to match the top-level element of your component
+export interface CountdownProps extends ComponentProps<'div'> {
+  /**
+   * The date the countdown should end
+   */
+  endDate: string;
```
scottdickerson, line 10: I think it's probably best to take a `Date` type here since otherwise you'd have to worry about inconsistencies in parsing based on different formats

src/components/Countdown/Countdown.tsx
```diff
@@ -0,0 +1,151 @@
...
+   */
+  intervalDescription?: string;
+  /**
+   * Text translation for days
+   */
+  daysLabel?: string;
+  /**
+   * Text translation for hours
+   */
+  hoursLabel?: string;
```
scottdickerson, line 22: if you take the locale as a parameter, you could just use this I believe to format the words for Hours, Minutes, Days, and Seconds: https://date-fns.org/v4.1.0/docs/formatDistanceStrict

src/components/Countdown/Countdown.tsx
```diff
@@ -0,0 +1,151 @@
...
+    const calculateTimeLeft = useCallback(() => {
+      const targetTime = new Date(endDate).getTime();
+      const currentTime = new Date().getTime();
+      const difference = targetTime - currentTime;
+
+      // Adjust for GMT offsets and Daylight Saving Time
+      const targetDate = new Date(endDate);
+      const currentDate = new Date();
+
+      const targetOffset = targetDate.getTimezoneOffset();
```
scottdickerson, line 71: I think by changing to a Date as a param, you won't have to manually adjust. The `formatDistanceToStrict` function of date-fns will do the calculations for you as well I believe

src/components/Countdown/Countdown.tsx
```diff
@@ -0,0 +1,151 @@
...
+    const { className: baseClassName, ...commonProps } = getCommonProps(props, 'Countdown');
+
+    const calculateTimeLeft = useCallback(() => {
+      const targetTime = new Date(endDate).getTime();
+      const currentTime = new Date().getTime();
+      const difference = targetTime - currentTime;
+
+      // Adjust for GMT offsets and Daylight Saving Time
+      const targetDate = new Date(endDate);
+      const currentDate = new Date();
```
scottdickerson, line 69: For me, rather than have `timeLeft` be the state, it makes more sense to store `currentDateTime`, `setCurrentDateTime` as state, then have the `date-fns` calculate the time lefts from that.

src/components/Countdown/Countdown.tsx
```diff
@@ -0,0 +1,151 @@
...
+      const timer = setInterval(() => {
+        setTimeLeft(calculateTimeLeft());
+      }, 1000);
+
+      return () => clearInterval(timer);
+    }, [calculateTimeLeft, endDate]);
+
+    return (
+      <div {...commonProps} className={classnames(baseClassName, className)} {...props} ref={ref}>
+        <div className={`${baseClassName}__countdown-container`}>
```
scottdickerson, line 108: it looks like there is an aria-role for a timer: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/timer_role

src/components/Countdown/Countdown.tsx
```diff
@@ -0,0 +1,96 @@
...
+ * [Figma Link](https://www.figma.com/design/hMu9IWH5N3KamJy8tLFdyV/EASEL-Compendium%3A-Tokens%2C-Components-%26-Patterns?node-id=10850-35410&m=dev)
+ *
+ * [Storybook Link](https://phillips-seldon.netlify.app/?path=/docs/components-countdown--overview)
+ */
+const Countdown = forwardRef<HTMLDivElement, CountdownProps>(
+  (
+    {
+      endDateTime,
+      label = 'Lots Close in',
+      intervalDescription = 'Lots Close in 1-minute intervals',
```
sgrund14, line 47: was it 1 or 2 minute intervals?

sgrund14 (approved):
LGTM just one question about the interval

src/components/Countdown/Countdown.tsx
scottdickerson: supposedly 1 even though the figma says otherwise..

PR #398, opened 2024-10-24
feat(Tags): L3-3816 Create Tags Component

zhenpanda:
**Jira ticket**

(https://phillipsauctions.atlassian.net/browse/L3-3816)

**Screenshots**

Screenshot 2024-10-24 112259
https://github.com/user-attachments/assets/8b5a1b66-f6dd-44ba-b70f-ff6534133f46

**Summary**
Tags are the filters on the sale page. These will populate based on the Artists or Makers in the sale.
They are not buttons but have buttons as (x) on them that can be removed by click or focus+enter key.

Add high-level overview of the changes that were made and explanation of WHY they were made

**Regression Test**

**Evidence of testing**

- Post logs, screenshots, etc

New Components


src/components/Tags/Tags.tsx
```diff
@@ -0,0 +1,113 @@
...
+  /**
+   * Tag item label.
+   */
+  label: string;
+}
+
+const Tags = ({ id, index, className, onRemove, label }: TagsProps) => {
+  return (
+    <div className={classnames(`${px}-tag`, `${px}-button`, className)} aria-label="Close Tag">
+      <>
```
scottdickerson, line 62: I don't believe you need this fragment @zhenpanda

src/components/Tags/Tags.tsx
```diff
@@ -0,0 +1,113 @@
...
+        <div className={`${px}-tag-label`}>{label}</div>
+        <div
+          onClick={() => onRemove(label)}
+          className={`${px}-close-button`}
+          data-testid={`${id}-item-${index}-close-button`}
+        >
+          <IconButton className={`${px}-tag-x`}>
```
scottdickerson, line 69: could we use `tag__close`

src/components/Tags/Tags.tsx
```diff
@@ -0,0 +1,113 @@
...
+  return (
+    <div
+      className={classnames(`${px}-${type}`, baseClassName, className)}
+      {...commonProps}
+      {...props}
+      data-testid={`${type}-${id}`}
+    >
+      {tagsList.map((tag, index) => {
+        return (
+          <Tags label={tag} onRemove={onRemove} key={`${tag}-key`} className={tagClassName} id={id} index={index} />
```
scottdickerson, line 92: I'm wondering if we need `index` at all

src/components/Tags/Tags.tsx
```diff
@@ -0,0 +1,113 @@
...
+        );
+      })}
+      {tagsList.length > 0 && (
+        <Button
+          onClick={onClear}
+          data-testid={`${id}-clear-all-button`}
+          className={`${px}-${type}-clear-all-button`}
+          aria-label="Clear All"
```
scottdickerson, line 100: this should be a translatable string property

src/components/Tags/Tags.tsx
```diff
@@ -0,0 +1,113 @@
...
+    <div className={classnames(`${px}-tag`, `${px}-button`, className)} aria-label="Close Tag">
+      <>
+        <div className={`${px}-tag-label`}>{label}</div>
+        <div
+          onClick={() => onRemove(label)}
+          className={`${px}-close-button`}
+          data-testid={`${id}-item-${index}-close-button`}
+        >
+          <IconButton className={`${px}-tag-x`}>
+            <span>X</span>
```
scottdickerson, line 70: this seems taller than the Asset in Figma.  Could we use the actual icon asset?

scottdickerson: export { default as Close } from './close.svg?react';

src/components/Tags/_tags.scss
```diff
@@ -0,0 +1,113 @@
...
+  gap: $spacing-sm;
+
+  .#{$px}-tag {
+    background-color: $pure-white;
+    border: 1px solid $button-hover;
+    border-radius: 6.25rem;
+    cursor: default;
+    gap: 0;
+    height: 2rem;
+    padding: 0 0.25rem 0 0.875rem;
```
scottdickerson, line 15: it seems like the padding on the tag is using spacing tokens in the Figma 
```
display: flex;
height: var(--Design-Cheats-for-Button-Tertiary---Fixed-Height, 28px);
padding: var(--Spacers-Nil, 0px) var(--Spacers-X-Small, 8px) var(--Spacers-Nil, 0px) var(--Spacers-Small, 12px);
justify-content: center;
align-items: center;
gap: var(--Spacers-Nil, 0px);
```

src/components/Tags/_tags.scss
```diff
@@ -0,0 +1,113 @@
...
+        color: $pure-black;
+      }
+    }
+  }
+}
+
+.#{$px}-tags-list-clear-all-button {
+  border-color: $pure-white;
+  display: flex;
+  gap: 0;
```
scottdickerson, line 57: is seems like there is a gap but no padding on clear-all
```
display: flex;
height: var(--Design-Cheats-for-Button-Tertiary---Fixed-Height, 28px);
padding: var(--Spacers-Nil, 0px);
justify-content: center;
align-items: center;
gap: var(--Spacers-Micro, 4px);
```

src/components/Tags/_tags.scss
```diff
@@ -0,0 +1,113 @@
...
+        outline-offset: none;
+      }
+    }
+  }
+
+  .#{$px}-label {
+    color: $pure-black;
+    font-size: 0.875rem;
+    font-variation-settings: 'wght' 600;
+    padding-right: 0.5rem;
```
scottdickerson, line 111: I don't see this in the Figma

src/components/Tags/Tags.tsx
```diff
@@ -0,0 +1,113 @@
...
+          onClick={onClear}
+          data-testid={`${id}-clear-all-button`}
+          className={`${px}-${type}-clear-all-button`}
+          aria-label="Clear All"
+          variant={ButtonVariants.tertiary}
+        >
+          <div className={`${px}-left-arrow`}>
+            <ChevronRight />
+          </div>
+          <div className={`${px}-label`}>Clear All</div>
```
scottdickerson, line 106: This needs to be the same translatable string

src/components/Tags/Tags.tsx
```diff
@@ -0,0 +1,113 @@
...
+   */
+  className?: string;
+  /**
+   * Individual Tag class
+   */
+  tagClassName?: string;
+  /**
+   * List of tags to render
+   */
+  tagsList: string[];
```
scottdickerson, line 25: Can this be a ReactNode, and can we just use a composable interface, so no `tagsList` prop, just use the Children and pass each tag as Children.  if so you won't need to pass the `onRemove` here nor the `tagClassName`

src/components/Tags/Tags.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+const Tags = ({ id, index, className, onRemove, label }: TagsProps) => {
+  return (
+    <div className={classnames(`${px}-tag`, `${px}-button`, className)} aria-label="Close Tag">
+      <>
+        <div className={`${px}-tag-label`}>{label}</div>
+        <div
+          onClick={() => onRemove(label)}
+          className={`${px}-close-button`}
+          data-testid={`${id}-item-${index}-close-button`}
```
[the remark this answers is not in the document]
zhenpanda: Yea the id is there for the test cases so it could select a specific button.

src/components/Tags/Tags.tsx
zhenpanda: This is added for unit testing

src/components/Tags/_tags.scss
zhenpanda: This is to overwrite the CSS of children inside of the Button

src/components/Tags/Tags.tsx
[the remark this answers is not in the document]
scottdickerson: couldn't the testcase just select using the id, and not need the index?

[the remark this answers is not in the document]
zhenpanda: ya i'll make the changes to remove index from the tags component

src/components/Tags/Tags.tsx
```diff
@@ -0,0 +1,109 @@
...
+export const Tags = ({ id, className, onRemove, label }: TagsProps) => {
+  return (
+    <div className={classnames(`${px}-tag`, `${px}-button`, className)} aria-label="Close Tag">
+      <div className={`${px}-tag__label`}>{label}</div>
+      <div
+        onClick={() => onRemove(label)}
+        className={`${px}-close-button`}
+        data-testid={`${id}-item-close-button`}
+      >
+        <IconButton className={`${px}-close-button__close`}>
```
davidicus, line 69: I think this class should reference its parent base class. It identifies the element and styles as part of the tag. 
```suggestion
        <IconButton className={`${px}-tag__close-button`}>
```

src/components/Tags/Tags.tsx
```diff
@@ -16,32 +21,28 @@ export interface TagsListProps {
...
   /**
-   * List of tags to render
-   */
-  tagsList: string[];
-  /**
-   * `onRemove` callback removes the clicked tag from the list
+   * Individual Tag class
    */
-  onRemove: (tag: string) => void;
+  tagClassName?: string;
```
scottdickerson, line 30: now that you have children, you no longer need this prop as they can just pass it directly to the Tag component child

src/components/Tags/Tags.tsx
```diff
@@ -16,32 +21,28 @@ export interface TagsListProps {
...
   /**
    * `onRemove` callback provides the tag that is to be removed as a string
    */
   onClear: () => void;
+  /**
+   * Clear All label is translatable
+   */
+  i18n?: I18nObject;
```
scottdickerson, line 38: I've seen this pattern before, but we haven't been doing this in other components, instead just saying `clearAllLabel`

src/components/Tags/Tags.tsx
```diff
@@ -56,29 +57,28 @@ export interface TagsProps {
...
+        <IconButton className={`${px}-close-button__close`}>
+          <CloseIcon />
+        </IconButton>
+      </div>
     </div>
   );
 };
 
-const TagsList = ({ className, tagClassName, onRemove, onClear, tagsList = [], ...props }: TagsListProps) => {
+const TagsList = ({ className, children, tagClassName, i18n = { clearAllLabel: 'Clear All' }, onClear, ...props }: TagsListProps) => {
```
scottdickerson, line 77: the `i18n` default here actually won't work well because it the user ever passes an `i18n` object (for other future labels) WITHOUT a `clearAllLabel` prop, it will override and leave it undefined.

src/components/Tags/Tags.tsx
```diff
@@ -0,0 +1,109 @@
...
+      >
+        <IconButton className={`${px}-close-button__close`}>
+          <CloseIcon />
+        </IconButton>
+      </div>
+    </div>
+  );
+};
+
+const TagsList = ({ className, children, tagClassName, i18n = { clearAllLabel: 'Clear All' }, onClear, ...props }: TagsListProps) => {
```
davidicus, line 77: Since this is a single string that we need for tag list should the `i18n` prop be swapped out for a `clearAllLabel` prop?

src/components/Tags/Tags.tsx
davidicus, line 69: The close icons are not matching the figma size. SHould the IconButton have size `sm`?

src/components/Tags/_tags.scss
```diff
@@ -0,0 +1,134 @@
...
+      height: $button-tertiary-height;
+      justify-content: center;
+      padding: 0;
+
+      &__close {
+        font-size: 0.75rem;
+        height: 1.65rem;
+        width: 1.65rem;
+
+        svg {
```
scottdickerson, line 40: the size of the icon doesn't match the Figma at 16px x 16px

src/components/Tags/_tags.scss
```diff
@@ -0,0 +1,134 @@
...
+      border-color: $button-hover;
+      border-radius: 6.25rem;
+      gap: 0;
+      outline-offset: 0;
+      padding: 0;
+    }
+
+    .#{$px}-left-arrow {
+      height: 1.25rem;
+      transform: scaleX(-1);
```
scottdickerson, line 103: could we change this to a rotate instead, I find scaleX negative to be confusing

src/components/Tags/_tags.scss
```diff
@@ -0,0 +1,134 @@
...
+    &:focus-visible {
+      border-color: $button-hover;
+      border-radius: 6.25rem;
+      gap: 0;
+      outline-offset: 0;
+      padding: 0;
+    }
+
+    .#{$px}-left-arrow {
+      height: 1.25rem;
```
scottdickerson, line 102: In the figma the icon size is 16px, but it's using some kind of weird snowflake token var: `--design-cheats-for-button-small-icon-size-11`.  We need to check with @RyanOlanoPHILLIPS to see how this scales at different breakpoints.  I assume it's not supposed to stay 16 the whole time

src/components/Tags/_tags.scss
```diff
@@ -0,0 +1,134 @@
...
+          outline-offset: none;
+        }
+      }
+    }
+
+    .#{$px}-label {
+      color: $pure-black;
+      font-size: 0.875rem;
+      font-variation-settings: 'wght' 600;
+      padding-right: 0.5rem;
```
scottdickerson, line 130: I don't see any `padding-right` in the Figma for this label

src/components/Tags/Tags.tsx
zhenpanda: the `onClick` is on a div that's nested in `Tags` how I can tell it which element the `onRemove` function should be attached to?

scottdickerson:
> the clear all button is not vertically centered next to the tags:
> 
> image. I think you need an `align items: center` on the tags-list Flexbox div

@zhenpanda this is still misaligned for me: 
image

scottdickerson (approved):
LGTM

src/components/Tags/Tags.tsx
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * ## Overview
+ *
+ * Overview of Tags component
+ *
+ * [Figma Link](https://www.figma.com/design/hMu9IWH5N3KamJy8tLFdyV/EASEL-Compendium%3A-Tokens%2C-Components-%26-Patterns?node-id=11648-3225&node-type=canvas&m=dev)
+ *
+ * [Storybook Link](https://phillips-seldon.netlify.app/?path=/docs/components-tags--overview)
+ */
+const TagsList = ({ className, children, clearAllLabel = 'Clear All', onClear, ...props }: TagsListProps) => {
```
scottdickerson, line 78: @zhenpanda I'm not sure this has been created by the `npm run create-component` script since it doesn't have the `forwardRef` to the tags list.  Could you use that going forward for new components?

scottdickerson (changes requested):
Could you add the `forwardRef` like our component template does?

scottdickerson (approved):
LGTM

PR #401, opened 2024-10-24
feat(header): L3-4294 links underline fix

constantinehuzenko:
**Jira ticket**

[Jira Ticket ID](https://phillipsauctions.atlassian.net/browse/L3-4294)

**Screenshots**
| Before | After |
| ------ | ----- |
| image | image |

**Summary**

Fixed the link underline

- src/scss/_type.scss: updated the underline mixin

- test that links underline looks correct

**Regression Test**

**Evidence of testing**

- Post logs, screenshots, etc

New Components

### DIFF
--- a/src/scss/_type.scss
+++ b/src/scss/_type.scss
@@ -37,9 +37,10 @@
   font-family: Distinct, sans-serif;
 }
 
-@mixin underline($padding: 0, $width: 0.0625rem, $color: $pure-black) {
-  border-bottom: $width solid $color;
-  padding-bottom: $padding;
+@mixin underline($offset: 8px, $thickness: 1px, $color: $pure-black) {
+  text-decoration: underline;
+  text-decoration-thickness: $thickness;
+  text-underline-offset: $offset;
 }
 
 @mixin hText(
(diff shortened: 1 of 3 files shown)


scottdickerson (approved):
LGTM, this is great @constantinehuzenko

src/scss/_type.scss
```diff
@@ -37,9 +37,10 @@
   font-family: Distinct, sans-serif;
 }
 
-@mixin underline($padding: 0, $width: 0.0625rem, $color: $pure-black) {
-  border-bottom: $width solid $color;
-  padding-bottom: $padding;
+@mixin underline($offset: 8px, $thickness: 1px, $color: $pure-black) {
```
scottdickerson, line 40: should the offset be a spacing token by default? I suspect so because as the text gets larger, the offset needs to increase

scottdickerson (dismissed):
one small change needed

src/scss/_type.scss
constantinehuzenko: image
image
image

still looks good on different size

davidicus (approved):

scottdickerson (commented):
LGTM

PR #404, opened 2024-10-28
fix(lotdetails): L3-3327 minor changes to input and accordion to match compendium

sgrund14:
**Jira ticket**

[L3-3327](https://phillipsauctions.atlassian.net/browse/L3-3327)

**Screenshots**
**Before**
Screenshot 2024-10-28 at 10 30 36 AM
**After**
Screenshot 2024-10-28 at 10 31 56 AM

**Summary**

Match invalid email styles to compendium

- keep invalid text black, increase font weight of input text, red outline on focused invalid input
- add onOpen/onClose props to accordion item

**Regression Test**

**Evidence of testing**

- Post logs, screenshots, etc

New Components

[L3-3327]: https://phillipsauctions.atlassian.net/browse/L3-3327?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Input/_input.scss
+++ b/src/components/Input/_input.scss
@@ -15,8 +15,10 @@ $lg: #{$px}-input--lg;
 
   &__label {
     color: $pure-black;
+    display: flex;
     font-size: 0.75rem;
     font-variation-settings: 'wght' 600;
+    gap: 0.25rem;
     margin-bottom: 0.5rem;
     word-break: break-word;
 
@@ -30,8 +32,13 @@ $lg: #{$px}-input--lg;
     border: 1px solid $keyline-gray;
     border-radius: 0.1875rem;
     font-size: 0.8125rem;
+    font-variation-settings: 'wght' 600;
     margin-bottom: 0.25rem;
     padding: $padding-xsm;
+
+    &::placeholder {
+      font-variation-settings: 'wght' 400;
+    }
 
     &:focus-within {
       outline: 1px solid $pure-black;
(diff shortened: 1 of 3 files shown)


src/components/Input/_input.scss
```diff
@@ -15,8 +15,10 @@ $lg: #{$px}-input--lg;
 
   &__label {
     color: $pure-black;
+    display: flex;
     font-size: 0.75rem;
     font-variation-settings: 'wght' 600;
+    gap: 0.25rem;
```
scottdickerson, line 21: I see a gap of spacer-xsm in Figma in the parent div

src/components/Input/_input.scss
```diff
@@ -30,9 +32,14 @@ $lg: #{$px}-input--lg;
     border: 1px solid $keyline-gray;
     border-radius: 0.1875rem;
     font-size: 0.8125rem;
+    font-variation-settings: 'wght' 600;
     margin-bottom: 0.25rem;
     padding: $padding-xsm;
```
scottdickerson, line 37: seems like the bottom padding to the Error is padding-micro in the Figma?

sgrund14: sure, I can add in

this would stay the same though, since it's not the error?

sgrund14: also this padding value is for the inner padding of the input

davidicus (approved):
LGTM. The padding thing is minor IMO and does not really align to a defined token. I have left a comment on the Figma to see if the difference is intentional.

src/components/Input/_input.scss
sgrund14: Screenshot 2024-10-30 at 11 00 20 AM

🤔

sgrund14: maybe some of that space is coming from the line height of the input

sgrund14:
hm, publish job failing...

PR #408, opened 2024-10-29
feat(header): L3-4326 updated spacing and search click

constantinehuzenko:
**Jira ticket**

[L3-4326](https://phillipsauctions.atlassian.net/browse/L3-4326)

**Screenshots**
| Before | After |
| ------ | ----- |
| image | image |

**Summary**

Remove margin from menu item, fixed logo margin, fixed auction links (need to update remix to see change), updated search click

**Regression Test**

**Evidence of testing**

- Post logs, screenshots, etc

New Components

### DIFF
--- a/src/components/Link/_link.scss
+++ b/src/components/Link/_link.scss
@@ -2,6 +2,7 @@
 
 .#{$px}-link {
   color: $pure-black;
+  display: inline;
   text-decoration: none;
 
   &--email,
--- a/src/components/Navigation/NavigationList/_navigationList.scss
+++ b/src/components/Navigation/NavigationList/_navigationList.scss
@@ -7,7 +7,6 @@ $top-transition: top 0.3s cubic-bezier(0.4, 0, 0.2, 1);
   flex-direction: column;
   justify-content: left;
   opacity: 1;
-  padding: $spacing-md 0;
 
   &--offscreen {
     opacity: 0;
--- a/src/components/Search/Search.test.tsx
+++ b/src/components/Search/Search.test.tsx
@@ -28,6 +28,13 @@ describe('Search component', () => {
     await userEvent.click(button);
     expect(searchInput).toHaveFocus();
   });
+  it('should open input once clicked on input container', async () => {
+    const { container } = renderWithContext(<Search />);
+    const inputContainer = container.getElementsByClassName('seldon-search__container__inner');
+    const searchInput = screen.getByTestId('search-input');
+    await userEvent.click(inputContainer[0]);
+    expect(searchInput).toHaveFocus();
+  });
   it('should close form when close button is clicked', async () => {
     renderWithContext(<Search />);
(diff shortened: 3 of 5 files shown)


adietrich3074 (commented):
Why are we removing the space between the language selector and the main nav list? I remember being explicitly told to add this in. Was the figma for the mobile nav updated?

constantinehuzenko:
> Why are we removing the space between the language selector and the main nav list? I remember being explicitly told to add this in. Was the figma for the mobile nav updated?

we don't have it on Prod, now I'm thinking if we should have it there
image

adietrich3074 (approved):

src/components/Navigation/NavigationItem/_navigationItem.scss
```diff
@@ -69,6 +69,10 @@
     margin-bottom: 0;
   }
 
+  a {
+    display: inline;
```
davidicus, line 73: This should probably be inline-block. Unless  this is to shrink the clickable area

davidicus: Looking on mobile, I think we may want to add a media query to only apply this to desktop sizes. I think on mobile the expectation would be to have the whole menu bar be clickable, no?

davidicus:
As Alex mentioned, there is spacing in between the language selector and the last link as shown in the figma https://www.figma.com/design/hMu9IWH5N3KamJy8tLFdyV/EASEL-Compendium%3A-Tokens%2C-Components-%26-Patterns?node-id=10572-6185&node-type=canvas&m=dev

src/components/Navigation/NavigationItem/_navigationItem.scss
davidicus: @constantinehuzenko Was this added explicitly for mobile? I think the expectation on mobile would be that the whole line would be the link and not jsut the texts, because of how we have styled it. 

On desktop I would make it inline-block. Also the LI seems to be `inline`. That should be made to be inline-block too or removed all together .

constantinehuzenko:
> As Alex mentioned, there is spacing in between the language selector and the last link as shown in the figma https://www.figma.com/design/hMu9IWH5N3KamJy8tLFdyV/EASEL-Compendium%3A-Tokens%2C-Components-%26-Patterns?node-id=10572-6185&node-type=canvas&m=dev

Ryan approved that we shouldn't add space there

src/components/Navigation/NavigationItem/_navigationItem.scss
```diff
@@ -69,6 +69,12 @@
     margin-bottom: 0;
   }
 
+  a {
+    @include isHeaderMobile {
+      display: block;
```
davidicus, line 74: This take care of mobile but you still have the pointer on desktop when its not on the link. If we take the cursor pointer off the `li` it will just turn to pointer when on anchor.

davidicus (approved):

PR #418, opened 2024-11-06
feat(filter): create dumb filter component

Zach-OfAllTrades:
**Jira ticket**

[Jira Ticket ID](https://phillipsauctions.atlassian.net/browse/L3-4114)

**Screenshots**
image
image

https://github.com/user-attachments/assets/620b86de-3b0f-4236-988e-d14c24ef666c

**Summary**

This creates a dumb filter component but also manages the state between child and parent filters (like when clicking view all) along with the transition between the two. These components are meant to be composable. In storybook, I implemented this filter in a Drawer component as it will be on the sale page. 

- Added FilterControl, Filter, FilterHeader, and FilterValue components. 

Being that this is a dumb component, the actual filtering logic will need to be written when implemented. The main thing to test here is that the component and its children render correctly, view all appears after 10 filters is exceeded, that all filters are listed when view all is click, a back button is present on the view all, and that there is a transition between the two. 

New Components

### DIFF
--- a/src/components/Filter/Filter.stories.tsx
+++ b/src/components/Filter/Filter.stories.tsx
@@ -1,0 +1,71 @@
+import { Meta } from '@storybook/react';
+import Filter from './Filter';
+import FilterHeader from './FilterHeader';
+import FilterValue from './FilterValue';
+
+// More on how to set up stories at: https://storybook.js.org/docs/react/writing-stories/introduction
+const meta = {
+  title: 'Components/Filter',
+  component: Filter,
+} satisfies Meta<typeof Filter>;
+
+export default meta;
+
+type FilterDimension = { label: string; disabled?: boolean | undefined };
+
+type FilterType = {
+  label: string;
+  filterDimensions: FilterDimension[];
+};
+
+type PropTypes = {
+  filter: FilterType;
+  viewAllLimit: number;
+  isLast: boolean;
+  isViewingAll: boolean;
+};
+
+const filter: FilterType = {
+  label: 'Artists & Makers',
+  filterDimensions: [
+    { label: 'Jimmy' },
+    { label: 'Bob' },
+    { label: 'Alan' },
+    { label: 'Nick' },
+    { label: 'Joe' },
+    { label: 'Fred' },
+    { label: 'Rob' },
+    { label: 'Roy' },
+    { label: 'disabled', disabled: true },
+  ],
+};
+
+export const Playground = (props: PropTypes) => {
+  const { filter, viewAllLimit, isLast, isViewingAll } = props;
+  return (
+    <Filter key={filter.label} isLast={isLast} viewAllLimit={viewAllLimit} viewingAll={isViewingAll}>
+      <FilterHeader label={filter.label} />
+      {filter.filterDimensions.map((value: FilterDimension) => (
+        <FilterValue
+          key={value.label}
+          label={value.label}
+          onChange={(e) => {
+            e;
+          }}
+          inputType="checkbox"
+          disabled={value?.disabled}
+        />
+      ))}
+    </Filter>
+  );
+};
+
+// More on writing stories with args: https://storybook.js.org/docs/react/writing-stories/args
+Playground.args = {
+  filter,
+  viewAllLimit: 10,
+  isLast: true,
+  isViewingAll: false,
+};
+
+Playground.argTypes = {};
--- a/src/components/Filter/Filter.tsx
+++ b/src/components/Filter/Filter.tsx
@@ -1,0 +1,85 @@
+import {
+  ComponentProps,
+  forwardRef,
+  Children,
+  cloneElement,
+  isValidElement,
+  ReactNode,
+  Dispatch,
+  SetStateAction,
+} from 'react';
+import { getCommonProps, px } from '../../utils';
+import classnames from 'classnames';
+import FilterHeader, { FilterHeaderProps } from './FilterHeader';
+import { FilterValueProps } from './FilterValue';
+import Button from '../Button/Button';
+import { ButtonVariants } from '../Button/types';
+import ChevronNextIcon from '../../assets/chevronNext.svg?react';
+
+// You'll need to change the ComponentProps<"htmlelementname"> to match the top-level element of your component
+export interface FilterProps extends ComponentProps<'div'> {
+  children: ReactNode;
+
+  // If true, do not include bottom border. True by default, do not define if you do not wish to use separators.
+  isLast?: boolean;
+
+  // Setter for values to display in view all
+  setViewAllFilter?: Dispatch<SetStateAction<ReactNode[]>>;
+
+  // Number of values to display before truncating with view all button
+  viewAllLimit?: number;
+
+  // Whether this is a view all filter or not
+  viewingAll?: boolean;
+}
+/**
+ * ## Overview
+ *
+ * Component to display filtering dimensions
+ *
+ * [Figma Link](https://www.figma.com/design/OvBXAq48blO1r4qYbeBPjW/RW---Sale-Page-(PLP)?node-id=892-71019&node-type=frame&t=AsBDn4UgUEjNUnf7-0)
+ *
+ * [Storybook Link](https://phillips-seldon.netlify.app/?path=/docs/components-filter--overview)
+ */
+const Filter = forwardRef<HTMLDivElement, FilterProps>(
+  ({ className, children, isLast = true, viewAllLimit = 10, viewingAll = false, setViewAllFilter, ...props }, ref) => {
+    const { className: baseClassName, ...commonProps } = getCommonProps(props, 'Filter');
+    const childrenArray = Children.toArray(children);
+
+    return (
+      <div
+        {...commonProps}
+        className={classnames(baseClassName, className, {
+          [`${px}-has-separators`]: !isLast,
+        })}
+        {...props}
+        ref={ref}
+      >
+        {childrenArray.map((child, index) =>
+          isValidElement(child)
+            ? child.type === FilterHeader
+              ? cloneElement(child, { showBack: viewingAll, setViewAllFilter } as FilterHeaderProps)
+              : cloneElement(child, { isHidden: !viewingAll && index > viewAllLimit } as FilterValueProps)
+            : child,
+        )}
+        {childrenArray.length > viewAllLimit && !viewingAll ? (
+          <Button
+            variant={ButtonVariants.tertiary}
+            onClick={() => {
+              setViewAllFilter && setViewAllFilter(childrenArray);
+            }}
+          >
+            {`View All`}
+            <ChevronNextIcon className={`${baseClassName}__chevron`} />
+          </Button>
+        ) : null}
+      </div>
+    );
+  },
+);
+
+Filter.displayName = 'Filter';
+
+export type FilterComponent = ReturnType<typeof Filter>;
+
+export default Filter;
--- a/src/components/Filter/FilterHeader.tsx
+++ b/src/components/Filter/FilterHeader.tsx
@@ -1,0 +1,51 @@
+import { ComponentProps, Dispatch, SetStateAction, forwardRef } from 'react';
+import { getCommonProps } from '../../utils';
+import classnames from 'classnames';
+import { Text, TextVariants } from '../Text';
+import Button from '../Button/Button';
+import { ButtonVariants } from '../Button/types';
+import ChevronNextIcon from '../../assets/chevronNext.svg?react';
+
+export interface FilterHeaderProps extends ComponentProps<'div'> {
+  // Text to display as the header
+  label: string;
+
+  // Whether the show all back button should be displayed (when viewing all filters)
+  showBack?: boolean;
+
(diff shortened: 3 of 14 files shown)


constantinehuzenko:
when you click on this part it should trigger checkbox
image

constantinehuzenko:
I think this after click state is kinda weird, maybe we can add some border radius?
image

constantinehuzenko:
the text's height is 8px which kinda cuts it, I'm afraid we will have incorrect spacing because of that 
image

src/components/Filter/Filter.tsx
```diff
@@ -0,0 +1,85 @@
...
+const Filter = forwardRef<HTMLDivElement, FilterProps>(
+  ({ className, children, isLast = true, viewAllLimit = 10, viewingAll = false, setViewAllFilter, ...props }, ref) => {
+    const { className: baseClassName, ...commonProps } = getCommonProps(props, 'Filter');
+    const childrenArray = Children.toArray(children);
+
+    return (
+      <div
+        {...commonProps}
+        className={classnames(baseClassName, className, {
+          [`${px}-has-separators`]: !isLast,
```
constantinehuzenko, line 53: not really clear, why do we need this separator?

constantinehuzenko: okay I see now, maybe we should `Separator` component for that, @scottdickerson wdyt?

constantinehuzenko: image

constantinehuzenko:
this one is responsible for showing that button, I guess it should be separated component, @scottdickerson please confirm
image

src/components/Filter/Filter.stories.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const { filter, viewAllLimit, isLast, isViewingAll } = props;
+  return (
+    <Filter key={filter.label} isLast={isLast} viewAllLimit={viewAllLimit} viewingAll={isViewingAll}>
+      <FilterHeader label={filter.label} />
+      {filter.filterDimensions.map((value: FilterDimension) => (
+        <FilterValue
+          key={value.label}
+          label={value.label}
+          onChange={(e) => {
+            e;
```
constantinehuzenko, line 53: not sure why we need this `e` here

src/components/Filter/Filter.tsx
```diff
@@ -0,0 +1,85 @@
...
+        {...props}
+        ref={ref}
+      >
+        {childrenArray.map((child, index) =>
+          isValidElement(child)
+            ? child.type === FilterHeader
+              ? cloneElement(child, { showBack: viewingAll, setViewAllFilter } as FilterHeaderProps)
+              : cloneElement(child, { isHidden: !viewingAll && index > viewAllLimit } as FilterValueProps)
+            : child,
+        )}
```
constantinehuzenko, line 64: can we move this conditions to variable? so we can reduce cognitive complexity of this part

src/components/Filter/Filter.tsx
```diff
@@ -0,0 +1,85 @@
...
+        {childrenArray.length > viewAllLimit && !viewingAll ? (
```
constantinehuzenko, line 65: can we move this one to variable as well?

src/components/Filter/Filter.tsx
```diff
@@ -0,0 +1,85 @@
...
+          <Button
+            variant={ButtonVariants.tertiary}
+            onClick={() => {
+              setViewAllFilter && setViewAllFilter(childrenArray);
```
constantinehuzenko, line 69: use `setViewAllFilter?.(childrenArray)`

src/components/Filter/Filter.tsx
```diff
@@ -0,0 +1,85 @@
...
+            }}
+          >
+            {`View All`}
```
constantinehuzenko, line 72: do we need this {``}?

constantinehuzenko:
I guess only one radio button can be selected

image

src/components/Filter/Filter.tsx
Zach-OfAllTrades: When there are multiple filters, the design shows there is a separator between them. This Filter component, for the majority of cases I would think, would be implemented alongside other Filters. I created the FilterControl component to wrap and manage these.

Zach-OfAllTrades:
> this one is responsible for showing that button, I guess it should be separated component, @scottdickerson please confirm image

I kept this as the same component because the only difference was number of options and button placement. When implemented inside the FilterControl component, the different states of being viewAll or not are managed and is displayed with a slide transition.

Zach-OfAllTrades:
> I guess only one radio button can be selected
> 
> image

I noticed this too but figured it would be handled when the logic was implemented but now on second thought I will probably need to pass some sort of isActive prop down...

scottdickerson:
> I think this after click state is kinda weird, maybe we can add some border radius? image

I'm not sure we want a "focus" state instead of a "focus-visible" state.  I don't think a border should show on an input when it's been clicked.

scottdickerson:
small alignment issue when I'm viewing All 

image

src/components/Filter/FilterValue.tsx
```diff
@@ -0,0 +1,59 @@
...
+ */
+const FilterValue = forwardRef<HTMLDivElement, FilterValueProps>(
+  ({ className, label, inputType, isHidden = false, onChange, disabled, ...props }, ref) => {
+    const { className: baseClassName, ...commonProps } = getCommonProps(props, 'FilterValue');
+    const disabledLabel = disabled ? 'disabled-label' : '';
+    return (
+      <>
+        {isHidden ? null : (
+          <div {...commonProps} className={classnames(baseClassName, className)} {...props} ref={ref}>
+            <Text variant={TextVariants.body2} className={`${baseClassName}__label ${disabledLabel}`}>
```
scottdickerson, line 38: we need a prefix here on disabledLabel to protect against css collisions


This should also be rendered as a label component with element="label" and ensure that the label is linked to the input for accessibility.  An example.  It might be better to actually use the `<Input label` instead of rendering your own text which hopefully should have the linkage built in and just apply custom styles to that label through scss.

<label for="tac">
  <input id="tac" type="checkbox" name="terms-and-conditions" />
  I agree to the Terms and Conditions.
</label>

src/components/Filter/Filter.tsx
```diff
@@ -0,0 +1,85 @@
...
+  isLast?: boolean;
+
+  // Setter for values to display in view all
+  setViewAllFilter?: Dispatch<SetStateAction<ReactNode[]>>;
+
+  // Number of values to display before truncating with view all button
+  viewAllLimit?: number;
+
+  // Whether this is a view all filter or not
+  viewingAll?: boolean;
```
scottdickerson, line 33: can this be `isViewingAll` so we follow the naming convention?

src/patterns/FilterControl/FilterControl.tsx
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * A container for filters that controls the state of parent and child filters
+ *
+ * [Figma Link](https://www.figma.com/design/OvBXAq48blO1r4qYbeBPjW/RW---Sale-Page-(PLP)?node-id=892-71019&node-type=frame&t=AsBDn4UgUEjNUnf7-0)
+ *
+ * [Storybook Link](https://phillips-seldon.netlify.app/?path=/docs/patterns-filtercontrol--overview)
+ */
+const FilterControl = forwardRef<HTMLDivElement, FilterControlProps>(({ className, children, ...props }, ref) => {
+  const { className: baseClassName, ...commonProps } = getCommonProps(props, 'FilterControl');
+  const [viewAllFilter, setViewAllFilter] = useState<ReactNode[]>([]);
```
scottdickerson, line 21: I think we might have to follow a different approach here to support Remix form

src/components/Filter/FilterHeader.tsx
```diff
@@ -0,0 +1,51 @@
...
+ * ## Overview
+ *
+ * The header of a filter
+ */
+const FilterHeader = forwardRef<HTMLDivElement, FilterHeaderProps>(
+  ({ className, label, showBack = false, setViewAllFilter, ...props }, ref) => {
+    const { className: baseClassName, ...commonProps } = getCommonProps(props, 'FilterHeader');
+
+    return (
+      <div {...commonProps} className={classnames(baseClassName, className)} {...props} ref={ref}>
```
scottdickerson, line 29: should this be a "Legend"? <legend>Choose your favorite monster</legend> https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset

src/components/Filter/Filter.tsx
```diff
@@ -0,0 +1,118 @@
...
+        className={classnames(baseClassName, className, {
+          [`${px}-has-separators`]: !isLast && !isViewingAll,
+          [`${px}-filter--hidden`]: isHidden,
+          'is-closing': isClosing,
+        })}
+        {...props}
+        ref={ref}
+      >
+        {/* TODO: REMOVE INLINE STYLING */}
+        <fieldset name={name} style={{ border: 'none', padding: 0 }}>
```
scottdickerson, line 95: were you planning on removing this as part of this?

src/patterns/FilterControl/FilterControl.tsx
```diff
@@ -0,0 +1,51 @@
...
+            isHidden: !isViewAllSet ? false : viewAllFilter !== childElement.props.name,
+            isViewingAll: isViewAllSet,
+            className: isViewAllSet && classnames(childElement.props.className, 'is-opening'),
+          } as FilterProps)
+        : childElement,
+    );
+
+    return (
+      <div {...commonProps} className={classnames(baseClassName, className)} {...props} ref={ref}>
+        <form action={action}>{parsedChildren}</form>
```
scottdickerson, line 43: I need to be able to pass in the `<form` element from outside this component so I can pass the Remix `<Form` component.  Perhaps following the `element` pattern we use in other places?

src/components/Filter/FilterValue.tsx
```diff
@@ -0,0 +1,90 @@
+import { ComponentProps, forwardRef } from 'react';
+import { getCommonProps, px } from '../../utils';
+import classnames from 'classnames';
+import { Text, TextVariants } from '../Text';
+import Input, { InputProps } from '../Input/Input';
+
+// You'll need to change the ComponentProps<"htmlelementname"> to match the top-level element of your component
+export interface FilterValueProps extends Omit<ComponentProps<'div'>, 'onChange' | 'value'> {
+  name: string;
```
scottdickerson, line 9: name needs to be passed in as Input props

scottdickerson: so does `onChange`

src/components/Filter/FilterValue.tsx
```diff
@@ -0,0 +1,90 @@
...
+  // Should this filter be hidden, most common use will be for view all truncation
+  isHidden?: boolean;
+
+  // Whether or not this filter is disabled
+  disabled?: boolean;
+
+  // Function to trigger when the state of this filter is changed
+  onChange: (e: React.ChangeEvent<HTMLInputElement>) => void;
+
+  inputProps?: InputProps;
```
scottdickerson, line 29: this is a full list of the props that are returned from our form library and will be in the `inputProps` field:
```
export type UseFormRegisterReturn<TFieldName extends InternalFieldName = InternalFieldName> = {
    onChange: ChangeHandler;
    onBlur: ChangeHandler;
    ref: RefCallBack;
    name: TFieldName;
    min?: string | number;
    max?: string | number;
    maxLength?: number;
    minLength?: number;
    pattern?: string;
    required?: boolean;
    disabled?: boolean;
};
```
I think to stop confusion you might just remove `inputProps` and just make all of those `top-level` props with the addition of ones that are not in that response

src/components/Filter/FilterValue.tsx
```diff
@@ -0,0 +1,90 @@
...
+    const { className: baseClassName, ...commonProps } = getCommonProps(props, 'FilterValue');
+    const disabledClass = disabled ? `${baseClassName}-disabled-label` : '';
+    return (
+      <div
+        {...commonProps}
+        className={classnames(baseClassName, className, { [`${px}-input__label--hidden`]: isHidden })}
+        {...props}
+        ref={ref}
+      >
+        <Text
```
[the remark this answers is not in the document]
Zach-OfAllTrades: Tried that originally but couldn't see where I had control over the placement, it was always right of the input. Could be missing something, though.

constantinehuzenko:
can we add padding bottom?

image

src/components/Filter/FilterValue.tsx
```diff
@@ -0,0 +1,76 @@
...
+ */
+const FilterValue = forwardRef<HTMLDivElement, FilterValueProps>(
+  (
+    {
+      className,
+      // name,
+      // label,
+      inputType = 'checkbox',
+      isHidden = false,
+      // onChange,
```
constantinehuzenko, line 31: let's remove unnecessary props

constantinehuzenko (approved):
LGTM

PR #421, opened 2024-11-11
feat(ContentPeek): L3-4299 - add new prop for specifying the minHeightThreshold for showing Peek controls

aahill50:
**Jira ticket**

[L3-4299](https://phillipsauctions.atlassian.net/browse/L3-4299)

**Screenshots**
Before:
| Before | After |
| --- | --- |
|<video src="https://github.com/user-attachments/assets/32cfcabc-2d2c-4e4b-b08b-6e727ec761f0">|<video src="https://github.com/user-attachments/assets/1448bed9-96b9-4dfd-a216-2c38a5300d3c">|

**Summary**

The requirement for L3-4299 specifies that any content under 38rem should be shown without Peek controls, but any content above 38 rem should show the Peek, with the peeked content maxed at 30rem. With the current component this isn't psosible, so I added a new prop: `minHeightThreshold`. By default, it will be the same as the `maxHeight`, so it will work the same as it does today if `minHeightThreshold` isn't specified, but it can be passed separately to enable us to set different thresholds for when the Peek controls should be shown. high-level overview of the changes that were made and explanation of WHY they were made. I also added a `heightUnits` prop that can be used to specify whether the numbers passed in are in `px` or `rem`

**Regression Test**

**Evidence of testing**

- Post logs, screenshots, etc

New Components

[L3-4299]: https://phillipsauctions.atlassian.net/browse/L3-4299?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/ContentPeek/ContentPeek.tsx
+++ b/src/components/ContentPeek/ContentPeek.tsx
@@ -6,6 +6,7 @@ import { ButtonVariants } from '../Button/types';
 import { Collapsible, CollapsibleContent, CollapsibleTrigger } from '../Collapsible';
 import PlusIcon from '../../assets/plus.svg?react';
 import MinusIcon from '../../assets/minus.svg?react';
+import { HeightUnits } from './utils';
 
 export interface ContentPeekProps extends React.HTMLAttributes<HTMLDivElement> {
   /**
@@ -32,6 +33,10 @@ export interface ContentPeekProps extends React.HTMLAttributes<HTMLDivElement> {
    * Used to set a minimum height for the content before enabling the Peek functionality. Defaults to the same value as maxHeight if not provided.
    */
   minHeightThreshold?: number;
+  /**
+   * The unit used for setting height values. Defaults to 'px'.
+   */
+  heightUnits?: HeightUnits;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


src/components/ContentPeek/ContentPeek.tsx
```diff
@@ -28,6 +29,14 @@ export interface ContentPeekProps extends React.HTMLAttributes<HTMLDivElement> {
    * Maximum height of the content when collapsed
    */
   maxHeight?: number;
+  /**
+   * Used to set a minimum height for the content before enabling the Peek functionality. Defaults to the same value as maxHeight if not provided.
+   */
+  minHeightThreshold?: number;
```
sgrund14, line 35: in what case is this and max height different?

aahill50: The requirements in the ticket are for the Peek to only occur if the content height is > 38rem, but the Peek itself needs to be 30rem. This is to prevent the situation where the content is only a little bit bigger than the size of the Peek, and clicking the "Read More" button would only reveal a few pixels worth of content

PR #424, opened 2024-11-15
feat(select): L3-3818 updated select

constantinehuzenko:
**Jira ticket**

[L3-3818](https://phillipsauctions.atlassian.net/browse/L3-3818)

**Screenshots**
image
image

**Summary**

Added new variant and custom icon to the Select component

- src/components/Select/Select.stories.tsx: added new story
- src/components/Select/Select.tsx: added new variant and custom icon

- verify that http://localhost:6006/?path=/story/components-select--tertiary has new component
- verify new select has not border
- verify that new select's arrows responses on open event

**Regression Test**

**Evidence of testing**

- Post logs, screenshots, etc

New Components

[L3-3818]: https://phillipsauctions.atlassian.net/browse/L3-3818?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -1,15 +1,23 @@
 import * as React from 'react';
 import classnames from 'classnames';
-
 import { px, useNormalizedInputProps } from '../../utils';
-
 import { InputProps } from '../Input/Input';
+import './select.scss';
+import { SelectVariants } from './types';
 
 export interface SelectProps extends InputProps {
   /**
    * Option elements that are selectable
    */
   children: React.ReactNode;
+  /**
+   * Determines if you want to show the icon
+   */
+  showCustomIcon?: boolean;
+  /**
+   * Determines the variant of the select
+   */
+  variant: SelectVariants;
 }
 
 /**
@@ -31,6 +39,8 @@ const Select = React.forwardRef(
       disabled,
       hideLabel,
       id,
+      showCustomIcon = false,
+      variant = SelectVariants.default,
       inline,
       invalid,
       invalidText,
@@ -48,6 +58,9 @@ const Select = React.forwardRef(
   ) => {
     const type = 'select';
     const inputProps = useNormalizedInputProps({ disabled, id, invalid, invalidText, readOnly, type, warn, warnText });
+    const [isOpen, setIsOpen] = React.useState(false);
+    const handleIsOpen = () => setIsOpen((prev) => !prev);
+    const closeDropdown = () => setIsOpen(false);
 
     const wrapperClassnames = classnames(`${px}-${type}-input`, `${px}-input`, `${px}-input--${size}`, {
       [`${px}-input--inline`]: inline,
@@ -58,19 +71,32 @@ const Select = React.forwardRef(
       [`${className}__wrapper`]: className,
     });
 
+    const selectClassnames = classnames(`${px}-input__input`, {
+      className,
+      [`${px}-input__select--open`]: isOpen && showCustomIcon,
+      [`${px}-input__select--closed`]: !isOpen && showCustomIcon,
+      [`${px}-input__select--tertiary`]: variant === SelectVariants.tertiary,
+    });
+
+    const handleClick = (e: React.MouseEvent<HTMLSelectElement>) => {
+      handleIsOpen();
+      onClick?.(e);
+    };
+
     return (
       <div className={wrapperClassnames}>
         <label htmlFor={id} className={classnames(`${px}-input__label`, { [`${px}-input__label--hidden`]: hideLabel })}>
           {labelText}
         </label>
         <select
-          className={classnames(`${px}-input__input`, { className })}
+          className={selectClassnames}
           data-testid={id}
           defaultValue={defaultValue}
           disabled={inputProps.disabled}
           id={id}
           onChange={onChange}
-          onClick={onClick}
+          onBlur={closeDropdown}
+          onClick={handleClick}
           ref={ref}
           value={value}
           {...rest}
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -58,9 +58,6 @@ const Select = React.forwardRef(
   ) => {
     const type = 'select';
     const inputProps = useNormalizedInputProps({ disabled, id, invalid, invalidText, readOnly, type, warn, warnText });
-    const [isOpen, setIsOpen] = React.useState(false);
-    const handleIsOpen = () => setIsOpen((prev) => !prev);
-    const closeDropdown = () => setIsOpen(false);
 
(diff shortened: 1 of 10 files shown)


src/components/Select/_select.scss
```diff
@@ -0,0 +1,40 @@
...
+        &:focus-visible {
+          background-color: $pure-white;
+          border-radius: 0;
+          outline-color: $soft-black;
+          outline-offset: 4.5px;
+        }
+      }
+
+      &--closed {
+        background: url("data:image/svg+xml,<svg height='16' width='16' viewBox='0 0 16 16' fill='none' xmlns='http://www.w3.org/2000/svg'><path fill-rule='evenodd' clip-rule='evenodd' d='M8.06566 9.12623L12.2097 4.98219C12.3561 4.83574 12.7442 4.80954 12.9673 5.0327C13.1905 5.25585 13.1643 5.64386 13.0178 5.79031L8.42162 10.3865C8.34739 10.4607 8.21109 10.5041 8.06566 10.4997C7.92022 10.5041 7.78392 10.4607 7.70969 10.3865L3.1135 5.79031C2.96705 5.64386 2.94085 5.25585 3.16401 5.0327C3.38716 4.80954 3.77517 4.83574 3.92162 4.98219L8.06566 9.12623Z' fill='black'/></svg>")
```
sgrund14, line 23: should this icon just be saved in the icons directory?

src/components/Select/Select.tsx
```diff
@@ -1,15 +1,22 @@
...
 
 export interface SelectProps extends InputProps {
   /**
    * Option elements that are selectable
    */
   children: React.ReactNode;
+  /**
+   * Determines if you want to show the icon
+   */
+  showCustomIcon?: boolean;
```
sgrund14, line 15: NBD but naming is somewhat confusing, since you can't pass in your own custom icon. I was almost expecting another prop called `customIcon` which let me pass in my own icon component

could be something like, `showArrowIcon` or `showCaretIcon` or `showIcon` or something like that

scottdickerson: IMHO the custom Icon should always render in place of the default Select appearance as the Easel compendium has it there @constantinehuzenko  and the default HTML `appearance: none` should be used for that Select component

src/components/Select/Select.tsx
```diff
@@ -58,19 +70,32 @@ const Select = React.forwardRef(
...
         <select
-          className={classnames(`${px}-input__input`, { className })}
+          className={selectClassnames}
           data-testid={id}
           defaultValue={defaultValue}
           disabled={inputProps.disabled}
           id={id}
           onChange={onChange}
-          onClick={onClick}
+          onBlur={closeDropdown}
```
scottdickerson, line 97: isn't opening and closing handled already by the standard select HTML code.  Why do we need to have our own `isOpen` state locally here in the Select component?

scottdickerson: is it possible to apply our custom icon styles using a CSS attribute selector on the Select.  For instance `:focus` pseudo selector.  As much as possible we don't want any internal state in our Seldon components

src/components/Select/_select.scss
```diff
@@ -0,0 +1,40 @@
+@use '#scss/allPartials' as *;
+
+.#{$px}-select-input {
+  select.#{$px}-input {
+    &__select {
+      &--tertiary {
```
scottdickerson, line 6: awesome, thank you

src/components/Select/Select.tsx
constantinehuzenko: @scottdickerson we need it for icon's rotation, html's select has not any `open` state or attribute

src/components/Select/Select.tsx
scottdickerson, line 15: @constantinehuzenko  I still think this this component should be opinionated and force the custom icon to be used (since otherwise the UX is wrong)

scottdickerson: I don't think there's a usecase where a consumer would want to use the default `<select` appearance for the caret

src/components/Select/_select.scss
```diff
@@ -0,0 +1,40 @@
...
+
+      &--open {
+        background: url("data:image/svg+xml,<svg height='16' width='16' viewBox='0 0 16 16' fill='none' xmlns='http://www.w3.org/2000/svg'><path fill-rule='evenodd' clip-rule='evenodd' d='M8.06566 6.37378L12.2097 10.5178C12.3561 10.6643 12.7442 10.6905 12.9673 10.4673C13.1905 10.2442 13.1643 9.85614 13.0178 9.70969L8.42162 5.1135C8.34739 5.03927 8.21109 4.99593 8.06566 5.00031C7.92022 4.99593 7.78392 5.03927 7.70969 5.1135L3.1135 9.70969C2.96705 9.85614 2.94085 10.2442 3.16401 10.4673C3.38716 10.6905 3.77517 10.6643 3.92162 10.5178L8.06566 6.37378Z' fill='black'/></svg>")
+          no-repeat;
+      }
+
+      &--closed,
+      &--open {
+        appearance: none;
+        background-position: calc(100% - 0.75rem) center;
```
scottdickerson, line 35: from what I can tell the padding on the right doesn't adjust at the different breakpoints so at smaller ones, it doesn't look great: 
image

scottdickerson: image

scottdickerson (changes requested):
a few small changes

src/scss/_vars.scss
```diff
@@ -328,6 +334,8 @@ $snowflake-menu-padding: 1.25rem;
     --spacing-lg: 3rem;
     --spacing-xl: 4rem;
     --spacing-xxl: 5rem;
+    --select-icon-padding: 1.75rem;
```
sgrund14, line 337: why have these in vars rather than directly in the select styles? seems a little further removed than necessary

src/components/Select/Select.tsx
```diff
@@ -31,6 +39,8 @@ const Select = React.forwardRef(
       disabled,
       hideLabel,
       id,
+      showIcon = true,
```
sgrund14, line 42: will this affect language selector?

sgrund14 (approved):
minor comments but non-blocking

src/components/Select/Select.tsx
constantinehuzenko: nope

src/scss/_vars.scss
constantinehuzenko: fair enough, it was late night decision

scottdickerson:
seeing some strangeness here: 

https://github.com/user-attachments/assets/bcdda646-30d5-4ab5-988b-4743f820bc49

PR #425, opened 2024-11-15
feat(objecttile): L3-4346 - create object tile pattern

davidicus:
**Jira ticket**

[L3-4346](https://phillipsauctions.atlassian.net/browse/L3-4346)

**Summary**

Create Object tile pattern based off [this](https://www.figma.com/design/hMu9IWH5N3KamJy8tLFdyV/EASEL-Compendium%3A-Tokens%2C-Components-%26-Patterns?node-id=10914-34792&node-type=canvas&m=dev) figma.

To support this I have updated the countdown and detail component.

- Created all files for ObjectTile pattern
- Created all files BidSnapshot pattern
- Add props to Detail component
- Add prop to countdown.
- added test to detail and countdown
- Changed enum being referenced by SaleHeaderBanner to Match the enum we are going to get back from the API
- Updated story and test for SaleHeaderBanner

- ObjectTile matches Figma

**Regression Test**

- DetailList, Detail, Countdown, and SaleHeaderBanner still match production

**Evidence of testing**

image
image

New Components

[L3-4346]: https://phillipsauctions.atlassian.net/browse/L3-4346?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/assets/PhillipsLogo.svg
constantinehuzenko, line 1: why did we change it?

src/components/Countdown/Countdown.tsx
```diff
@@ -76,13 +81,17 @@ const Countdown = forwardRef<HTMLDivElement, CountdownProps>(
       <div {...commonProps} className={classnames(baseClassName, className)} {...props} ref={ref}>
         <div className={`${baseClassName}__countdown-container`} role="timer" aria-label={label}>
           <span>{label}</span>
-          {timeLeft.days > 0 ? <Duration duration={timeLeft} unit="days" locale={dateFnsLocale} /> : undefined}
+          {timeLeft.days > 0 ? (
+            <Duration duration={timeLeft} unit="days" locale={dateFnsLocale} formatDurationStr={formatDurationStr} />
+          ) : undefined}
```
constantinehuzenko, line 86: I think we are using null instead of undefined for conditional rendering, even though undefined is allowed with new react version, null is better for our consistency

constantinehuzenko:
I guess we will have jump content between different states
https://github.com/user-attachments/assets/6b6e8696-81a7-499e-b214-5f2c90e00a83

src/patterns/BidSnapshot/BidSnapshot.tsx
```diff
@@ -0,0 +1,147 @@
...
+    },
+    ref,
+  ) => {
+    const { className: baseClassName, ...commonProps } = getCommonProps(props, 'BidSnapshot');
+    const currentDate = new Date();
+    return (
+      <div {...commonProps} className={classnames(baseClassName, className)} {...props} ref={ref}>
+        {
+          // auctionStatus is not past (auction is active or upcoming),
+          !(auctionStatus === AuctionStatus.past) ? (
```
constantinehuzenko, line 94: can we avoid nested conditions?

I think cleaner way to create more single conditions like `condition ? <Component/> : null`
pretty sure it will be easier to read

constantinehuzenko: or we can move them to variables

src/patterns/BidSnapshot/BidSnapshot.tsx
```diff
@@ -0,0 +1,147 @@
...
+              <DetailList hasSeparators className={`${baseClassName}__text`}>
+                <Detail label={startingBidText} value={bids[0]} hasWrap={false} />
+              </DetailList>
+            ) : (
+              // otherwise, show current bid
+              <DetailList hasSeparators className={`${baseClassName}__text`}>
+                <Detail
+                  label={currentBidText}
+                  subLabel={bids.length > 0 && `(${bidsLabelText(bids.length)})`}
+                  value={bids[bids.length - 1]}
```
constantinehuzenko, line 107: `bids[bids.length - 1]` can be a variable, you use it 4-5 times

constantinehuzenko:
I think spacing is off

image
image

constantinehuzenko:
don't see this icon 
image

constantinehuzenko:
current status and current bid are bold in figma

image
image

constantinehuzenko:
can you add second state to the story? or the way to switch to that story

image

src/patterns/ObjectTile/_objectTile.scss
```diff
@@ -0,0 +1,96 @@
...
+      gap: 0 $spacing-sm;
+
+      .#{$px}-detail__value {
+        text-align: unset;
+      }
+    }
+  }
+
+  &__estimate {
+    // display: flex;
```
constantinehuzenko, line 82: can we remove? if we need it let's comment why we need it

constantinehuzenko (changes requested):
Minor questions about code, but no blocker
UI spacing, missing icon and bold text are necessary

davidicus:
> I guess we will have jump content between different states
> 
>  Recording.2024-11-15.163340.mp4

This video does not play

src/patterns/BidSnapshot/BidSnapshot.tsx
sgrund14: agree that nested ternaries can get freaky

src/patterns/ObjectTile/ObjectTile.tsx
```diff
@@ -0,0 +1,174 @@
...
+      modelText,
+      referenceNumber,
+      titleText,
+      withdrawnText,
+      ...props
+    },
+    ref,
+  ) => {
+    const { className: baseClassName, ...commonProps } = getCommonProps(props, 'ObjectTile');
+    Element = (props.href ? 'a' : 'div') as ElementType<
```
sgrund14, line 88: will `element` prop ever get used if passed in? seems like it can only ever be `a` or `div`. maybe this should be `props.href ? 'a' : Element` and `Element` defaults to `div`

src/patterns/ObjectTile/ObjectTile.tsx
```diff
@@ -0,0 +1,174 @@
...
+      Omit<ObjectTileProps, 'imageUrl' | 'lotNumber' | 'referenceNumber'> & { 'data-testid': string }
+    >;
+    return (
+      <Element {...commonProps} className={classnames(baseClassName, className)} {...props} ref={ref}>
+        {imageUrl ? (
+          <img
```
sgrund14, line 94: could use seldon Image component here potentially for loading & fallback states

src/patterns/ObjectTile/ObjectTile.tsx
```diff
@@ -0,0 +1,174 @@
...
+            className={`${baseClassName}__img`}
+            src={imageUrl}
+            alt={titleText ?? `image for lot number ${lotNumber}`}
+          />
+        ) : (
+          <div className={`${baseClassName}__img`}>
+            <PhillipsLogo />
```
sgrund14, line 101: I like this as a fallback, maybe it should go in the Seldon Image instead of "Error loading image" 🤔

src/patterns/ObjectTile/_objectTile.scss
```diff
@@ -0,0 +1,96 @@
...
+    // flex-wrap: wrap;
+    // justify-content: space-between;
+
+    // .#{$px}-object-tile__estimate__label {
+    //   font-variation-settings: "wght" 600;
+    //   margin-bottom: 0;
+    //   margin-right: 1rem;
+    // }
+
+    // .#{$px}-object-tile__estimate__value {
```
sgrund14, line 92: cleanup commented out styles?

src/patterns/BidSnapshot/BidSnapshot.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+        {
+          // Countdown timer
+          lotCloseDate && lotCloseDate > currentDate ? (
+            <Countdown
+              endDateTime={lotCloseDate}
+              label={closingText}
+              variant={CountdownVariants.compact}
+              locale={SupportedLanguages[lang]}
+              formatDurationStr={(str) => str.replace(/seconds?/, 'sec').replace(/minutes?/, 'min')}
```
scottdickerson, line 136: doesn't this need to handle translations for the different locales?

src/components/Countdown/Countdown.tsx
```diff
@@ -13,6 +13,10 @@ export interface CountdownProps extends ComponentProps<'div'> {
    * The date the countdown should end
    */
   endDateTime: Date;
+  /**
+   * Function to modify strings coming from date-fns (Minutes -> Mins etc)
+   */
+  formatDurationStr?: (duration: string) => string;
```
scottdickerson, line 19: instead of passing the actual string, it might be better to mirror the way that `formatDuration` does it: https://date-fns.org/v4.1.0/docs/formatDuration.  `duration: number, interval: `day | minute ...`. This would allow having the default translations for these durations returned by `date-fns` based on locale, then allow the custom shorter translation where necessary, since the string replace would be different for each translation.

src/patterns/BidSnapshot/BidSnapshot.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+export interface BidSnapshotProps extends ComponentProps<'div'> {
+  /**
+   * Active bid of current user. -  '$1,000'
+   */
+  activeBid?: string;
+  /**
+   * State of the object
+   */
+  auctionStatus?: AuctionStatus;
```
scottdickerson, line 19: I'm wondering if this particular component should be in `remix` because it could reuse the actual API `AuctionStateEnum`.

src/patterns/SaleHeaderBanner/SaleHeaderBanner.tsx
```diff
@@ -73,8 +73,8 @@ const SaleHeaderBanner = forwardRef<HTMLDivElement, SaleHeaderBannerProps>(
     ref,
   ) => {
     const { className: baseClassName, ...commonProps } = getCommonProps(props, 'SaleHeaderBanner');
-    const isOpenForBidding = auctionState === AuctionState.openForBidding;
-    const isClosed = auctionState === AuctionState.past;
+    const isOpenForBidding = auctionState === AuctionStatus.live;
+    const isClosed = auctionState === AuctionStatus.past;
```
scottdickerson, line 77: I could make the same argument here, that we might have this in Remix rather than seldon...

src/patterns/BidSnapshot/BidSnapshot.tsx
davidicus: our other location in short but for future translations we may want to. We can surface this prop to the Bidsnapshot component and let the user pass in the function.

src/patterns/SaleHeaderBanner/SaleHeaderBanner.tsx
davidicus: Not sure if having an enum is reason to move to remix. We can use the enum to pass in here. I am mirroring the values that will be passed back from the API for AuctionStatus so if they differ we will probably want to update that enum as well.

src/components/Countdown/Countdown.tsx
davidicus: This does allow for the strings to be returned from date-fns. This function are passed those strings via `durationParts` in the Duration component and then being returned with any changes the function wants to perform on them.

src/patterns/BidSnapshot/BidSnapshot.tsx
davidicus: I added this as a prop for BidSnapshot so it can be passed in at implementation. Defaulting to the strings transitions for english as seen in the design.

davidicus:
> can you add second state to the story? or the way to switch to that story
> 
> image

The [BidSnapshot](https://deploy-preview-425--phillips-seldon.netlify.app/?path=/story/patterns-bidsnapshot--playground) has stories that show all the variations.

sgrund14 (approved):
LGTM

davidicus:
@constantinehuzenko In the spirit  of your comment in our retro. I am going to merge this. Any other comments, lets put in a follow up ticket.

PR #426, opened 2024-11-19
fix(Detail): L3-4592 add widths to detail values at different breakpoints

aahill50:
**Jira ticket**

[L3-4592](https://phillipsauctions.atlassian.net/browse/L3-4592)

**Screenshots**
| Before | After |
| ------ | ----- |
| L3-4592 - before | L3-4592 after |

**Summary**

Adds a width to the Detail values to match the figma (https://www.figma.com/design/hxqgsE26wM7hII0WaUaDfF/RW---TIMED-Lot-Details-(PDP)?node-id=191-6168&node-type=instance&m=dev), including different values at different breakpoints

- Updated _detail.scss to add widths at different breakpoints for the detail value
- Updated story to change grid layout when resized

**Regression Test**

**Evidence of testing**

- Post logs, screenshots, etc

New Components

### DIFF
--- a/src/components/Detail/_detail.scss
+++ b/src/components/Detail/_detail.scss
@@ -30,5 +30,22 @@
 
   &__value {
     font-variation-settings: 'wght' 400;
+    width: 176px;
+
+    @include media($breakpoint-sm) {
+      width: 212px;
+    }
+
+    @include media($breakpoint-md) {
+      width: 176px;
+    }
+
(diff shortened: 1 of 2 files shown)


scottdickerson (approved):
Code looks good, I'm just confirming with design that this is what we want

scottdickerson:
according to @RyanOlanoPHILLIPS we don't want to chop it like this, and it was a bug in the Figma:

PR #435, opened 2024-11-25
feat(SaleHeaderBanner): L3-4648 add countdown to sale header banner f…

aahill50:
…or timed auctions

**Jira ticket**

[L3-4648](https://phillipsauctions.atlassian.net/browse/L3-4648)

**Screenshots**
| Before | After |
| ------ | ----- |
| L3-4648 before | L3-4648 after |
| L3-4648 mobile before | L3-4648 mobile after |

**Summary**

Adds the countdown timer to the top of Time Auction Sale pages

**Regression Test**

**Evidence of testing**

- Post logs, screenshots, etc

New Components

[L3-4648]: https://phillipsauctions.atlassian.net/browse/L3-4648?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/patterns/SaleHeaderBanner/SaleHeaderBanner.tsx
+++ b/src/patterns/SaleHeaderBanner/SaleHeaderBanner.tsx
@@ -81,13 +81,14 @@ const SaleHeaderBanner = forwardRef<HTMLDivElement, SaleHeaderBannerProps>(
     const { className: baseClassName, ...commonProps } = getCommonProps(props, 'SaleHeaderBanner');
     const isOpenForBidding = auctionState === AuctionStatus.live;
     const isClosed = auctionState === AuctionStatus.past;
+
     return (
       <div {...commonProps} className={classnames(baseClassName, className)} {...props} ref={ref}>
-        {isOpenForBidding && auctionEndTime && (
+        {isOpenForBidding && auctionEndTime ? (
           <div className={`${baseClassName}__stack__mobile-countdown`}>
             {<Countdown endDateTime={auctionEndTime} showBottomBorder={false} />}
           </div>
-        )}
+        ) : null}
         <SeldonImage
           aspectRatio="16/9"
           src={imageSrcUrl}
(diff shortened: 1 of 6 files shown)


src/patterns/SaleHeaderBanner/SaleHeaderBanner.tsx
```diff
@@ -75,41 +81,48 @@ const SaleHeaderBanner = forwardRef<HTMLDivElement, SaleHeaderBannerProps>(
...
+          <SeldonImage
+            aspectRatio="16/9"
+            src={imageSrcUrl}
+            alt={String(auctionTitle)}
+            objectFit="cover"
+            className={`${baseClassName}__image`}
+          />
+          <PageMargin className={`${baseClassName}__stack-wrapper`} {...commonProps} {...props} ref={ref}>
+            {showCountdown && (
+              <Countdown endDateTime={new Date(auctionEndTime)} className={`${baseClassName}__countdown`} />
```
davidicus, line 98: I believe they moved the countdown timer above the banner image on mobile. Not sure why Figma is showing both above and below. We can utilize the media query util we use in other components like footer.

aahill50: Weird, I think it was only on the top when I first started working on it, but now I see it in both places in figma... But I just pushed up a commit that moves it to the top and gets rid of the bottom border for the mobile layout

sgrund14 (approved):

src/patterns/SaleHeaderBanner/SaleHeaderBanner.tsx
```diff
@@ -86,7 +97,11 @@ const SaleHeaderBanner = forwardRef<HTMLDivElement, SaleHeaderBannerProps>(
         />
         <PageMargin className={`${baseClassName}__stack-wrapper`} {...commonProps} {...props} ref={ref}>
           <div className={`${baseClassName}__stack`}>
-            {isOpenForBidding && children}
+            {isOpenForBidding && auctionEndTime && (
```
sgrund14, line 100: apparently we prefer doing this
```
{isOpenForBidding && auctionEndTime ? (<Component />) : null}
```
but I think it's ok if there are no numbers involved
sometimes you get weird stuff where if `auctionEndTIme` was 0, it would just render a 0 on the page
but that shouldn't happen here

aahill50: That's usually how I prefer to do it, but I was just following the convention that was already in the file... but I just pushed up a refactor commit to change all the conditional rendering in this file to ternarys

PR #441, opened 2024-11-27
feat(carousel): L3-4045 carousel overflow and zoom/pan behavior

sgrund14:
**Jira ticket**

[L3-4045](https://phillipsauctions.atlassian.net/browse/L3-4045)

**Screenshots**

**Dot Overflow**
https://github.com/user-attachments/assets/b0f74550-9de5-4cc5-b05c-733c3fcf4a10

**Carousel when item is zoomed and at edge**

https://github.com/user-attachments/assets/d0be4a68-359d-445b-bbae-b4668af9d0b6

**Summary**

Add carousel dots overflow behavior

https://www.figma.com/design/hMu9IWH5N3KamJy8tLFdyV/EASEL-Compendium%3A-Tokens%2C-Components-%26-Patterns?m=auto&node-id=15894-140133&focus-id=15894-140065

And the ability to pan to the next carousel item when zoomed in and at the edge of an item

- Use intersection observer in CarouselDots & CarouselDot to track dot visibility and shrink/unshrink
- Add tracking to the PinchZoom component to detect edges
- Add stories which show how to use to enable pan-when-zoomed-to-edge behavior

**Regression Test**

**Evidence of testing**

- Post logs, screenshots, etc

New Components

[L3-4045]: https://phillipsauctions.atlassian.net/browse/L3-4045?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Carousel/CarouselDot.tsx
+++ b/src/components/Carousel/CarouselDot.tsx
@@ -1,0 +1,44 @@
+import classNames from 'classnames';
+import { useInView } from 'react-intersection-observer';
+
+type Props = {
+  isSelected: boolean;
+  onClick: () => void;
+  baseClassName: string;
+  scrollableContainerRef: React.RefObject<HTMLDivElement>;
+  onInViewChange: (inView: boolean) => void;
+  isShrinked: boolean;
+};
+
+export const CarouselDot = ({
+  isSelected,
+  onClick,
+  baseClassName,
+  scrollableContainerRef,
+  onInViewChange,
+  isShrinked,
+}: Props) => {
+  const { ref } = useInView({
+    threshold: 0,
+    root: scrollableContainerRef.current,
+    onChange(inView) {
+      onInViewChange(inView);
+    },
+  });
+
+  return (
+    <button
+      ref={ref}
+      role="button"
+      onClick={onClick}
+      className={classNames(`${baseClassName}-pagination-dot-container`)}
+    >
+      <span
+        className={classNames(`${baseClassName}-pagination-dot`, {
+          [`${baseClassName}-pagination-dot-selected`]: isSelected,
+          [`${baseClassName}-pagination-dot-shrink`]: isShrinked,
+        })}
+      />
+    </button>
+  );
+};
--- a/src/components/Carousel/CarouselDot.tsx
+++ b/src/components/Carousel/CarouselDot.tsx
@@ -7,7 +7,7 @@ type Props = {
   baseClassName: string;
   scrollableContainerRef: React.RefObject<HTMLDivElement>;
   onInViewChange: (inView: boolean) => void;
-  isShrinked: boolean;
+  variant?: 'sm' | 'md';
 };
 
 export const CarouselDot = ({
@@ -16,7 +16,7 @@ export const CarouselDot = ({
   baseClassName,
   scrollableContainerRef,
   onInViewChange,
-  isShrinked,
+  variant = 'md',
 }: Props) => {
   const { ref } = useInView({
     threshold: 0,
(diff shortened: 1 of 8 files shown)


src/components/Carousel/CarouselDot.tsx
```diff
@@ -0,0 +1,44 @@
+import classNames from 'classnames';
+import { useInView } from 'react-intersection-observer';
+
+type Props = {
+  isSelected: boolean;
+  onClick: () => void;
+  baseClassName: string;
+  scrollableContainerRef: React.RefObject<HTMLDivElement>;
+  onInViewChange: (inView: boolean) => void;
+  isShrinked: boolean;
```
scottdickerson, line 10: might be better to do a `variant` `sm` or `lg` here

src/components/Carousel/CarouselDot.tsx
```diff
@@ -0,0 +1,44 @@
+import classNames from 'classnames';
+import { useInView } from 'react-intersection-observer';
+
+type Props = {
```
davidicus, line 4: why the type and not an interface?

davidicus: should be an interface with comments so we get the documentation in storybook.

src/components/Carousel/CarouselDot.tsx
```diff
@@ -0,0 +1,44 @@
...
+  return (
+    <button
+      ref={ref}
+      role="button"
+      onClick={onClick}
+      className={classNames(`${baseClassName}-pagination-dot-container`)}
+    >
+      <span
+        className={classNames(`${baseClassName}-pagination-dot`, {
+          [`${baseClassName}-pagination-dot-selected`]: isSelected,
```
davidicus, line 38: BEM has a `--` for modifier.
```suggestion
          [`${baseClassName}-pagination-dot--selected`]: isSelected,
          [`${baseClassName}-pagination-dot--${variant}`]: variant,
```

src/components/Carousel/CarouselDot.tsx
```diff
@@ -0,0 +1,44 @@
...
+      onInViewChange(inView);
+    },
+  });
+
+  return (
+    <button
+      ref={ref}
+      role="button"
+      onClick={onClick}
+      className={classNames(`${baseClassName}-pagination-dot-container`)}
```
davidicus, line 34: BEM has a `__` for elements.

```suggestion
      className={classNames(`${baseClassName}-pagination-dot__container`)}
```

src/components/Carousel/_carousel.scss
```diff
@@ -60,24 +60,45 @@
...
+        display: none; /* Safari and Chrome */
+      }
     }
 
     &-dot-container {
       align-items: center;
       display: flex;
       height: 8px;
       justify-content: center;
+      min-width: 8px;
```
davidicus, line 83: Do we need this if we are explicitly setting the width?

davidicus (commented):
awesome work. Some very minor comments. I see there are pre-existing BEM issues which, for consistency, would like to correct but not necessary right now.

src/components/Carousel/CarouselDot.tsx
sgrund14: this is internal to to `CarouselDots`, which is the exposed to consumers, so there's not actually a storybook for it

i can update to interface tho

src/components/Carousel/_carousel.scss
sgrund14: there's this weird thing where the dots gets squashed without it 🤷‍♂️

PR #443, opened 2024-12-02
fix(bidsnapshot): L3-4378 if i have bid on a lot but lost it should say won for

scottdickerson:
**Jira ticket**

[Jira Ticket ID](https://phillipsauctions.atlassian.net/browse/TICKET-ID)

**Screenshots**
| Before | After |
| ------ | ----- |
| image | image |

**Summary**

BidSnapshot was showing sold for instead of won for when the user lost the bid, I also refactored the code for BidSnapshot to make it more understandable

- change(BidSnapshot): soldprice label uses wonForText if the user has ever bid.  Split the inner Detail determinations to separated `<Detail` lines

- Go to the BidSnapshot story, switch the AuctionStatus to Past.  You should see `Sold for` and the sold for price.  Switch the `bidStatus` to all of the statuses, you should see `Won for`.  Now remove the `soldPrice` and you should not see the soldPrice line at all, but if the user has bid and won (`bidStatus`) you should still see the won/lost message

**Regression Test**

**Evidence of testing**

- Post logs, screenshots, etc

New Components

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,17 @@
+## [1.97.1](https://github.com/PhillipsAuctionHouse/seldon/compare/v1.97.0...v1.97.1) (2024-11-27)
+
+
+### Bug Fixes
+
+* **footer:** L3-4461 Update casing on newsletter subscription ([#440](https://github.com/PhillipsAuctionHouse/seldon/issues/440)) ([84e5a98](https://github.com/PhillipsAuctionHouse/seldon/commit/84e5a98d5e20efef82aa9c8db40c759baf0b8e61))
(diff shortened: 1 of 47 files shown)


davidicus:
This seems to contradict the [compendium](https://www.figma.com/design/hMu9IWH5N3KamJy8tLFdyV/EASEL-Compendium%3A-Tokens%2C-Components-%26-Patterns?node-id=11184-50758&node-type=text&version-id=2150577446630699987&m=dev) for ObjectTile

davidicus (changes requested):
Is there somewhere else that has this documented?

PR #462, opened 2025-01-08
feat(text): L3-3958 grid centering and text alignment

sgrund14:
**Jira ticket**

[L3-3958(https://phillipsauctions.atlassian.net/browse/L3-3958)

**Summary**

Don't default center the text in a grid item, give text an alignment option

- Remove text-align in centered grid items (child text should handle that)
- Add text alignment option to Text

- Text & Grid should function as expected

**Regression Test**

**Evidence of testing**

- Post logs, screenshots, etc

New Components

### DIFF
--- a/src/components/GridItem/_gridItem.scss
+++ b/src/components/GridItem/_gridItem.scss
@@ -11,13 +11,11 @@
 
   &-align-center {
     grid-column: math.ceil(math.div($total-cols + 1 - $span, 2)) / span #{$span};
-    text-align: center;
   }
 }
 
 .#{$px}-grid-item {
   justify-self: center;
-  text-align: center;
   width: 100%;
 
   @for $i from 1 through 2 {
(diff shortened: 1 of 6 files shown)


src/components/GridItem/_gridItem.scss
```diff
@@ -11,13 +11,11 @@
 
   &-align-center {
     grid-column: math.ceil(math.div($total-cols + 1 - $span, 2)) / span #{$span};
-    text-align: center;
```
sgrund14, line 14: this and below is center aligning the text in every grid item by default

that should be in control of the text, not the grid item?

src/components/Text/utils.ts
```diff
@@ -22,3 +22,7 @@ export const determineDefaultTextElement = (variant: TextVariants = TextVariants
 
   return 'span';
 };
+
+export const determineTextAlignment = (alignment: TextAlignments = TextAlignments.center): string => {
```
davidicus, line 26: do we need to abstract this? Since we are passing a default to the align prop and this is specifically a text util may be easier to just put the return value as the class?

davidicus (approved):
I am not opposed to abstraction but seems like it would be clearer to just have the class on the item where that class is coming from.

PR #468, opened 2025-01-10
fix(textsymbol): L3-5094 normalize symbol size

scottdickerson:
**Jira ticket**

[L3-5094](https://phillipsauctions.atlassian.net/browse/L3-5094)

**Screenshots**
| Before | After |
| ------ | ----- |
| image | image |

**Summary**

Some of the symbols we support are way too small in our font, this fix tries to normalize their size better so they're readable

- change(TextSymbol): switch `children` prop to `symbols` array since I have to examine each one and adjust its size
- change(_textSymbol.scss): add 3 size classes
- change(utils.ts): helper function to figure out which symbol is which size.

- Check the text symbol story

**Regression Test**

**Evidence of testing**

- Post logs, screenshots, etc

New Components

[L3-5094]: https://phillipsauctions.atlassian.net/browse/L3-5094?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Text/Text.test.tsx
+++ b/src/components/Text/Text.test.tsx
@@ -26,7 +26,7 @@ describe('Text', () => {
     renderText({
       children: (
         <>
-          Hello World<TextSymbol variant={TextSymbolVariants.lotNumber}>sup</TextSymbol>
+          Hello World<TextSymbol variant={TextSymbolVariants.lotNumber} symbols={['sup']}></TextSymbol>
         </>
       ),
     });
--- a/src/components/TextSymbol/utils.ts
+++ b/src/components/TextSymbol/utils.ts
@@ -1,0 +1,27 @@
+import { TextSymbolSize } from './types';
+
+export const determineSymbolSize = (symbol: string) => {
+  switch (symbol) {
+    case '≌':
+    case '܀':
+    case '•':
+      return TextSymbolSize.small;
+    case '▼':
+    case '♠':
+    case '≠':
+      return TextSymbolSize.medium;
+    case '∞':
+    case 'Ω':
+    case 'Δ':
+    case '†':
+    case '◆':
+    case 'Σ':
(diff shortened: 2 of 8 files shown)


aahill50 (approved):
This looks like a pretty good solution to me, but I'm also interested in what @constantinehuzenko thinks

src/components/Text/Text.test.tsx
```diff
@@ -26,7 +26,7 @@ describe('Text', () => {
     renderText({
       children: (
         <>
-          Hello World<TextSymbol variant={TextSymbolVariants.lotNumber}>sup</TextSymbol>
+          Hello World<TextSymbol variant={TextSymbolVariants.lotNumber} symbols={['sup']}></TextSymbol>
```
constantinehuzenko, line 29: `</TextSymbol>` we don't need closing tag if we are not passing anything inside

src/components/TextSymbol/utils.ts
```diff
@@ -0,0 +1,27 @@
+import { TextSymbolSize } from './types';
+
+export const determineSymbolSize = (symbol: string) => {
```
constantinehuzenko, line 3: instead of switch cases we can add either object with symbol as a key and check if it's in that object or maybe array and use includes function to check for that one

+ we'll have one source of truth what is small medium and big sized icons

constantinehuzenko (approved):
Small recommendation regarding switch cases

src/components/TextSymbol/utils.ts
scottdickerson: This is a great idea in general @constantinehuzenko but in this case I can't think of another usecase for where that object/array would be used

PR #478, opened 2025-01-21
fix(search): L3-5101 search bar flashing and removes typed search

SRangineni88:
**Jira ticket**

[L3-5101](https://phillipsauctions.atlassian.net/browse/L3-5101)

**Screenshots**
| Before | After |
| ------ | ----- |
| Paste Before Image | Paste After Image |

**Summary**

Search Bar Flashing and Removes Typed Search

Added css transition on search to minimize the flickering
Removed logic to close search immediately on key press or on clicking the search result link.

**Regression Test**

**Evidence of testing**

- Post logs, screenshots, etc

New Components

[L3-5101]: https://phillipsauctions.atlassian.net/browse/L3-5101?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Search/SearchResults/SearchResults.tsx
```diff
@@ -48,7 +47,7 @@ const SearchResults = ({
                 <Link
                   href={result.url}
                   onKeyDown={onKeyDown}
-                  onClick={() => closeSearch(false)}
+                  // onClick={() => closeSearch(false)}
```
Zach-OfAllTrades, line 50: Do we want to remove this completely? Looks to only be commented out here. 

Also, how is this closed now, then? I see we are controlled the opacity now but does the search bar clear when a link is clicked or does it just stay open?

src/components/Search/Search.tsx
```diff
@@ -203,7 +201,6 @@ const Search = ({
               >
                 <li key="viewAllSearchResults" className={`${baseClassName}__result`}>
                   <Link
-                    onClick={() => showSearch(false)}
```
davidicus, line 206: I think this is closing that search bar when clicking on one of the links in the search dropdown. I am not sure we want to remove this.

src/components/Search/Search.tsx
```diff
@@ -96,7 +95,6 @@ const Search = ({
       e.preventDefault();
       if (value && value.length > 2) {
         const allResultsLink = encodeURLSearchParams(getAllResultsLink(value));
-        showSearch(false);
```
davidicus, line 99: same comment about this. I would think we would want the search to close when we hit enter.

Zach-OfAllTrades (approved):

davidicus (changes requested):
I would like @adietrich3074 to review before letting this go as he had added some of this to fix other bugs in phillips-public

adietrich3074 (changes requested):
I'm not sure if the removal of all the close search functionality is the solution for this. The close search functionality was added to handle bugs where a user selects a search result, navigates to the page, and then hits the back button, the search will remain open and show broken on the page. I think this issue can be solved with the solution we implemented in remix which is to debounce the results. This allows time before the search gets sent and wait for the user to finish typing as opposed to making a call to the search api every time a character is entered. I believe these repeated calls with the results returning quickly is the cause of the flashing. There is an example of how the debouncing is done in the remix project in Header.tsx

src/scss/_vars.scss
```diff
@@ -103,8 +103,8 @@ $snwHeaderLink: 'snwHeaderLink';
 $snwFlyoutLink: 'snwFlyoutLink';
 $snwHeadingHero1: 'snwHeadingHero1';
 $snwHeadingHero2: 'snwHeadingHero2';
-$text-tokens: $button, $link, $email, $label, $badge, $blockquote, $heading1, $heading2, $heading3, $heading4, $heading5,
-  $title1, $title2, $title3, $title4, $body1, $body2, $body3, $string1, $string2, $string3, $snwFlyoutLink,
+$text-tokens: $button, $link, $email, $label, $badge, $blockquote, $heading1, $heading2, $heading3, $heading4,
```
scottdickerson, line 106: can we undo this change?  I'm guessing something about your prettier config isn't matching our repo one

scottdickerson (changes requested):
I don't think this should be fixed in seldon.  If you look at the Remix use of the Search widget in production: https://www.phillips.com/about.  I don't see the behavior that is referenced in the defect, which makes me think that there must be some approach that can be mirrored in Phillips-Public to fix this.  @adietrich3074 can help out there I think.

scottdickerson:
https://github.com/user-attachments/assets/2677b4f5-b44a-450b-bd66-0bf951eacfd1

PR #484, opened 2025-01-23
fix(tabs): L3-4707 Remove Focus Border From Tabs on Click

adietrich3074:
**Jira ticket**

[L3-4707](https://phillipsauctions.atlassian.net/browse/L3-4707)

**Screenshots**
| Before | After |
| ------ | ----- |
| ea2f0ca4-68b3-4650-bfc9-4cf56f89a6db| Screenshot 2025-01-23 121134 |

**Summary**

 Fixed tabbing functionality and corrected focus styles

- Adjusted padding for tab content
- Added tabIndex to TabTrigger component

- Verify that outline only shows when tabbing through tabs and not on click of the tab

New Components

[L3-4707]: https://phillipsauctions.atlassian.net/browse/L3-4707?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Tabs/TabTrigger.tsx
+++ b/src/components/Tabs/TabTrigger.tsx
@@ -32,7 +32,6 @@ const TabTrigger = forwardRef<HTMLButtonElement, TabTriggerProps>(({ value, onCl
       value={value}
       className={`${baseClassName}__tabs-trigger`}
       onClick={() => onClick?.(value)}
-      tabIndex={0}
       ref={ref}
     >
       {children}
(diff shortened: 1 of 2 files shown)


scottdickerson (changes requested):
the keyboard navigation here isn't quite right.  Hitting tab again should not change to the next <Tab>

```
: When focus moves to the tabs-widget, the active tab-element gets focus. When the focus is in the tab-element, the focus moves to the next focusable item (so, not to the next tab). This can mean either item in the active tab panel or first thing outside the widget.
Left Arrow: When the focus is on the active tab-element, the focus moves to the next tab on the list. If on the last tab, focus moves to the first tab. If tabs are automatically activated, activates the focused tab.
Right Arrow: When the focus is on the active tab-element, the focus moves to the previous tab on the list. If on the first item, moves focus to the last tab. If tabs are automatically activated, activates the focused tab.
```

https://dev.to/eevajonnapanula/keyboard-accessible-tabs-with-react-5ch4

PR #517, opened 2025-02-18
fix(Pagination): L3-5318 Pagination Dot update to match iOS

zhenpanda:
**Jira ticket**

[L3-5318](https://phillipsauctions.atlassian.net/browse/L3-5318)

**Screenshots**

https://github.com/user-attachments/assets/14e56a3e-3f6e-4bb6-b399-d4842497e518

**Figma link**

[Figma Link](https://www.figma.com/design/hMu9IWH5N3KamJy8tLFdyV/EASEL-Compendium%3A-Tokens%2C-Components-%26-Patterns?node-id=5792-407&p=f&focus-id=4612-77094&m=dev)

**Summary**

Add high-level overview of the changes that were made and explanation of WHY they were made

**Regression Test**

**Evidence of testing**

- Post logs, screenshots, etc

New Components

[L3-5318]: https://phillipsauctions.atlassian.net/browse/L3-5318?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Carousel/CarouselDots.tsx
+++ b/src/components/Carousel/CarouselDots.tsx
@@ -29,6 +29,10 @@ export interface CarouselDotsProps extends ComponentProps<'div'> {
  * The dots animate left to right as the user navigates through the carousel.
  *
  */
+const dotWidth = 10;
+const dotGap = 12;
+const centerDotContainer = 11;
+
 const CarouselDots = forwardRef<HTMLDivElement, CarouselDotsProps>(
   ({ className, maxDots = 9, position = 'inline', numberOfSlides = 0, ...props }, ref) => {
     const { className: baseClassName, ...commonProps } = getCommonProps(props, 'CarouselDots');
@@ -46,7 +50,7 @@ const CarouselDots = forwardRef<HTMLDivElement, CarouselDotsProps>(
       if (scrollableContainerRef.current) {
         scrollableContainerRef.current.scrollTo?.({
           // 8px width + 12px gap
-          left: index * (8 + 12) - scrollableContainerRef.current.offsetWidth / 2 + 10, // Center dot in container
+          left: index * (dotWidth + dotGap) - scrollableContainerRef.current.offsetWidth / 2 + centerDotContainer, // Center dot in container
           behavior: 'smooth',
         });
       }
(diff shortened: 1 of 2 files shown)


scottdickerson:
i don't think the dots are big enough according to the Figma link.  The dot in the code is 6x6 
image
but the dot in the figma is 8x8.  
In addition the focus border seems to be a little larger in the code than the Figma:
image

scottdickerson (changes requested):
the dots seem too small compared to the Figma

zhenpanda:
@scottdickerson Updated the dot size now is 8x8 like Figma, the gap in Figma is 12px which match and looks right when the dot size is correct.

Zach-OfAllTrades (approved):

src/components/Carousel/_carousel.scss
```diff
@@ -79,17 +79,31 @@
   &-dot__container {
     align-items: center;
     display: flex;
-    height: 8px;
+    height: 10px;
     justify-content: center;
-    min-width: 8px;
+    margin: 2px;
```
sgrund14, line 84: why is this margin necessary?

src/components/Carousel/_carousel.scss
```diff
@@ -79,17 +79,31 @@
...
+    min-width: 10px;
     padding: 1px;
     scroll-snap-align: center;
-    width: 8px;
+    width: 10px;
```
sgrund14, line 88: there's two spots in `CarouselDots` that need to be adjusted for the new width

https://github.com/PhillipsAuctionHouse/seldon/blob/8c3760bce1849eb5606c4c2480a915e0f7ab0ac0/src/components/Carousel/CarouselDots.tsx#L48-L49

8 needs to be replaced with 10 (probably should just set this to a variable), and 10 needs to be replaced with 11. What this does is center the dot that has just been clicked to

and

https://github.com/PhillipsAuctionHouse/seldon/blob/8c3760bce1849eb5606c4c2480a915e0f7ab0ac0/src/components/Carousel/CarouselDots.tsx#L110-L111

replace 8 with 10

sgrund14 (approved):

sgrund14 (changes requested):
meant to request the above changes

src/components/Carousel/_carousel.scss
zhenpanda: without the margin the focus will get cropped 

Screenshot 2025-02-19 131322

zhenpanda: I update the gap to have 10px so ends up with 12px gap with the margin

scottdickerson (approved):
LGTM

src/components/Carousel/CarouselDots.tsx
```diff
@@ -29,6 +29,10 @@ export interface CarouselDotsProps extends ComponentProps<'div'> {
  * The dots animate left to right as the user navigates through the carousel.
  *
  */
+const dotWidth = 10;
+const dotGap = 12;
+const centerDotContainer = 11;
```
sgrund14, line 34: non-blocking this should probably just be `(dotWidth + dotGap) / 2`

sgrund14 (approved):

PR #522, opened 2025-02-20
feat(input): L3-5461 react node as input label

sgrund14:
**Jira ticket**

[L3-5461](https://phillipsauctions.atlassian.net/browse/L3-5461)

**Summary**

Allow the input to accept a react node as a label (for adding variable styles)
Screenshot 2025-02-20 at 10 26 08 AM

also, adding in storybook a11y add on https://storybook.js.org/docs/writing-tests/accessibility-testing#set-up-the-a11y-addon

which give you an overview of accessibility violations / passes
Screenshot 2025-02-24 at 10 56 44 AM

**Regression Test**

**Evidence of testing**

- Post logs, screenshots, etc

New Components

[L3-5461]: https://phillipsauctions.atlassian.net/browse/L3-5461?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Input/Input.tsx
+++ b/src/components/Input/Input.tsx
@@ -49,7 +49,7 @@ export interface InputProps extends Omit<React.ComponentProps<'input'>, 'size'> 
   /**
    * Text that will be read by a screen reader when visiting this control
    */
-  labelText: string;
+  labelText: React.ReactNode;
 
   /**
    * Optional `onChange` handler that is called whenever `<input>` is updated
--- a/src/components/Input/Input.tsx
+++ b/src/components/Input/Input.tsx
@@ -160,7 +160,7 @@ const Input = React.forwardRef(
     return (
       <div className={wrapperClassnames}>
         <label
-          data-testid={`${id}-label`}
+          data-testid={`label-${id || generatedId}`}
           htmlFor={id || generatedId}
           className={classnames(`${px}-input__label`, { [`${px}-input__label--hidden`]: hideLabel })}
         >
--- a/src/components/Input/Input.tsx
+++ b/src/components/Input/Input.tsx
@@ -160,6 +160,7 @@ const Input = React.forwardRef(
     return (
       <div className={wrapperClassnames}>
         <label
+          data-testid={`${id}-label`}
           htmlFor={id || generatedId}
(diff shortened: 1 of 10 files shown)


src/components/Input/Input.tsx
```diff
@@ -47,7 +47,7 @@ export interface InputProps extends Omit<React.ComponentProps<'input'>, 'size'>
   /**
    * Text that will be read by a screen reader when visiting this control
    */
-  labelText: string;
+  labelText: React.ReactNode;
```
scottdickerson, line 50: can you add a testcase and tweak a story to use ReactNode instead of just text?

scottdickerson (changes requested):
just testcase and story updates

scottdickerson:
@sgrund14 can you also make sure in the testcase with ReactNode that accessibility still works.  I.e. make sure the `getByRole` can find the right label for a complex label.

davidicus (commented):
To the point of a11y, this does allow for elements that may not be accessible or permitted content. Which we can control but we will have to be diligent

sgrund14:
> diligent

Maybe we can just make a new prop that's like, label decorator or something

scottdickerson:
> > diligent
> 
> Maybe we can just make a new prop that's like, label decorator or something
> 
> Or labelSecondaryText, which is just a string and gets wrapped in a span which removes the bold text

alternatively we could use the `ariaLabel` approach on the input instead of the labelledBy/child approach to associate the label with the input for accessibility `ariaLabel` will always be a string.  We still need the labelledBy/child approach to get the input to get focus when you click on the label

sgrund14:
> > > diligent
> > 
> > 
> > Maybe we can just make a new prop that's like, label decorator or something
> > Or labelSecondaryText, which is just a string and gets wrapped in a span which removes the bold text
> 
> alternatively we could use the `ariaLabel` approach on the input instead of the labelledBy/child approach to associate the label with the input for accessibility `ariaLabel` will always be a string. We still need the labelledBy/child approach to get the input to get focus when you click on the label

how complex are we expecting these labels to get? won't the screen reader just extract any text content out of them?

scottdickerson (approved):
LGTM

PR #525, opened 2025-02-20
fix(useId): L3-5569 replace useId function with a third-party library

scottdickerson:
**Jira ticket**

[L3-5569](https://phillipsauctions.atlassian.net/browse/L3-5569)

**Summary**

Removes use of the new `useId` function in React 18, and bring in a third-party library.

This pull request introduces the `uuid` library to replace the `useId` hook for generating unique IDs across multiple components. The most important changes include updating dependencies in `package.json` and modifying several components to use `uuidv4` for ID generation.

Dependency updates:

* [`package.json`](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L63-R64): Added `uuid` library as a dependency.

Component updates:

* [`src/components/Carousel/CarouselDots.tsx`](diffhunk://#diff-c5f308f41986c7c6bef156edf8646b9efa0173c83ab41afdd1bae6f7cd30bb7eL3-R4): Imported `uuidv4` from `uuid` and replaced `useId` with `uuidv4` for generating `componentId`. [[1]](diffhunk://#diff-c5f308f41986c7c6bef156edf8646b9efa0173c83ab41afdd1bae6f7cd30bb7eL3-R4) [[2]](diffhunk://#diff-c5f308f41986c7c6bef156edf8646b9efa0173c83ab41afdd1bae6f7cd30bb7eL39-R40)
* [`src/components/Input/Input.tsx`](diffhunk://#diff-0414a1d786fc54ea337f45079b8c5d47254051dfc247a8f39d0f4c474fcb9f12R3): Imported `uuidv4` from `uuid` and replaced `useId` with `uuidv4` for generating `generatedId`. [[1]](diffhunk://#diff-0414a1d786fc54ea337f45079b8c5d47254051dfc247a8f39d0f4c474fcb9f12R3) [[2]](diffhunk://#diff-0414a1d786fc54ea337f45079b8c5d47254051dfc247a8f39d0f4c474fcb9f12L139-R140)
* [`src/components/Select/Select.tsx`](diffhunk://#diff-d6181d40b4597b102e57c0daacdac25ca9f5082357c4360133812235abc7bee1R6): Imported `uuidv4` from `uuid` and replaced `useId` with `uuidv4` for generating `generatedId`. [[1]](diffhunk://#diff-d6181d40b4597b102e57c0daacdac25ca9f5082357c4360133812235abc7bee1R6) [[2]](diffhunk://#diff-d6181d40b4597b102e57c0daacdac25ca9f5082357c4360133812235abc7bee1L62-R63)

- if testcases pass and phililips-public cn be upgraded we should be in good shpe.

**Regression Test**

**Evidence of testing**

- Post logs, screenshots, etc

New Components

[L3-5569]: https://phillipsauctions.atlassian.net/browse/L3-5569?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Carousel/CarouselDots.tsx
+++ b/src/components/Carousel/CarouselDots.tsx
@@ -1,6 +1,7 @@
 import classNames from 'classnames';
 import { EmblaCarouselType } from 'embla-carousel';
-import { ComponentProps, forwardRef, useCallback, useEffect, useState, useId, useRef, useMemo } from 'react';
+import { ComponentProps, forwardRef, useCallback, useEffect, useState, useRef, useMemo } from 'react';
+import { v4 as uuidv4 } from 'uuid';
 import { useCarousel } from './utils';
 import { getCommonProps } from '../../utils';
 import { CarouselDot } from './CarouselDot';
@@ -36,7 +37,7 @@ const centerDotContainer = 11;
 const CarouselDots = forwardRef<HTMLDivElement, CarouselDotsProps>(
   ({ className, maxDots = 9, position = 'inline', numberOfSlides = 0, ...props }, ref) => {
     const { className: baseClassName, ...commonProps } = getCommonProps(props, 'CarouselDots');
-    const componentId = useId();
+    const componentId = uuidv4();
     const { api, onSlideChange } = useCarousel();
     const [selectedIndex, setSelectedIndex] = useState(0);
     const [scrollSnaps, setScrollSnaps] = useState<number[]>(
--- a/package-lock.json
+++ b/package-lock.json
@@ -28,7 +28,8 @@
         "react-modal": "^3.16.1",
         "react-transition-group": "^4.4.5",
         "react-zoom-pan-pinch": "^3.6.1",
-        "usehooks-ts": "^3.1.0"
+        "usehooks-ts": "^3.1.0",
+        "uuid": "^11.1.0"
       },
       "devDependencies": {
         "@chromatic-com/storybook": "^3.2.3",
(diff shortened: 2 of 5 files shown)


src/components/Carousel/CarouselDots.tsx
```diff
@@ -36,7 +37,7 @@ const centerDotContainer = 11;
 const CarouselDots = forwardRef<HTMLDivElement, CarouselDotsProps>(
   ({ className, maxDots = 9, position = 'inline', numberOfSlides = 0, ...props }, ref) => {
     const { className: baseClassName, ...commonProps } = getCommonProps(props, 'CarouselDots');
-    const componentId = useId();
+    const componentId = uuidv4();
```
davidicus, line 40: Will this get recreated every time this component renders? I think this will give us hydration errors. 

Looking at the code below I see we are only using this for a key identifier, which react explicitly says not to do. The way it was defined looked to be static for the whole map. ie. creating one identifier on render and then passing that same id to each map loop. So should be the same of just using index here. 

I think we can just remove and it would be the same.

davidicus (commented):
@scottdickerson Looks like this did not resolve the issue on phillips-public. Since this may be problematic with our hydration errors and does not resolve the pp issue perhaps we scrap this PR?

src/components/Carousel/CarouselDots.tsx
scottdickerson: This change does work with React 17 actually but we'll cancel for now since there was an alternate fix for the problem.  The preferred strategy is to get phillips-public to upgrade to React 18

PR #566, opened 2025-03-26
feat(Sale and Lot Icons): L3-3417 icon updates

aahill50:
**Jira ticket**

[L3-3417](https://phillipsauctions.atlassian.net/browse/L3-3417)

**Screenshots**
Screenshot 2025-03-18 182802

**Summary**

Adds an `Icon` component to replace the various svgs and implementations we have across seldon and (soon) remix.

This pull request includes several changes to the `package.json` file and the addition of new SVG components in the `src/assets/formatted` directory. The goal is to introduce a new script for formatting SVG icons and provide documentation on how to use it.

### Summary of changes:

New scripts and dependencies:

* [`package.json`](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L40-R41): Added a new script `icons:format` to format SVG icons and added dependencies for `@svgr/cli`, `@svgr/plugin-jsx`, `@svgr/plugin-prettier`, and `@svgr/plugin-svgo`. [[1]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L40-R41) [[2]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R87-R90)

Documentation updates:

* [`src/assets/CONTRIBUTING.md`](diffhunk://#diff-f4f2d42b7710c4f72949aef87cadbd63c4b66d0e1cff2dcf54fae5487eab6552R1-R24): Added detailed instructions on how to convert SVG files into React components using the new `icons:format` script, including steps to ensure unique `clipPath` IDs and how to make icons recolorable.

New formatted SVG components added in `src/assets/formatted`

These changes improve the handling and usage of SVG icons within the project, making it easier to add and customize new icons.

**Regression Test**

**Evidence of testing**

- Post logs, screenshots, etc

New Components

[L3-3417]: https://phillipsauctions.atlassian.net/browse/L3-3417?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/assets/CONTRIBUTING.md
```diff
@@ -0,0 +1,24 @@
...
+- Open `NewIcon.jsx` and
+
+  - Make sure any clipPath ids used in the svg are unique. (This helps avoid issues that can occur when multiple svgs are used on the same page that share the same clipPath ids.) For example you can prepend the name of the component to the id. You will also need to update the corresponding element that references the id that you just changed. See the table below for an example:
+
+  - | Before Changing (Generic id: "a") | After Changing (Unique id: "NewIcon_a")   |
+    | --------------------------------- | ----------------------------------------- |
+    | `<clipPath id="a">...</clipPath>` | `<clipPath id="NewIcon_a">...</clipPath>` |
+    | `<g clipPath="url(#a)">...</g>`   | `<g clipPath="url(#NewIcon_a)">...</g>`   |
+
+  - If the icon needs to be recolorable, you will need to make a few more modifications to the jsx component. For svgs that don't need to be recolored (company logos for example), you can skip this step. Otherwise, open the jsx file, find any `fill` or `stroke` properties and set them to `fill={props.color}` or `stroke={props.color}`.
```
scottdickerson, line 20: is that really necessary?  Since it does a `{...props}` spread we could just pass the `fill` or `stroke` prop?

```
![image](https://github.com/user-attachments/assets/50abdc9f-24b1-4ce5-8448-219c7fb955e9)
```

package.json
```diff
@@ -37,7 +37,8 @@
     "lint": "tsc --noEmit && eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0 && npm run lint:styles && npm run lint:md",
     "lint:styles": "stylelint \"src/**/*.scss\" --fix",
     "lint:md": "markdownlint \"src/**/*.{md,mdx}\" --fix",
-    "prepare": "husky install"
+    "prepare": "husky install",
+    "icons:format": "npx @svgr/cli --out-dir src/assets/formatted --ignore-existing -- src/assets"
```
scottdickerson, line 41: we should use the typescript option so it generates the interfaces too [TypeScript](https://react-svgr.com/docs/options/#typescript)

src/assets/CONTRIBUTING.md
scottdickerson: looks like we might be able to use this instead to set currentColor? https://react-svgr.com/docs/options/#replace-attribute-value

package.json
scottdickerson: https://react-svgr.com/docs/options/#accessibility

scottdickerson: I don't see any harm in using these two options too: https://react-svgr.com/docs/options/#ref  https://react-svgr.com/docs/options/#memo might future proof us from other usecases or performance issues

scottdickerson (changes requested):
Requested some additional icon utility configuration

src/assets/CONTRIBUTING.md
aahill50: My idea was that the `color` prop would help abstract away most of the complexity of dealing with the underlying svg. We'd take the complexity hit up front when creating the component and determining exactly which parts need to support the passed in `color` prop, but then we end up with an `Icon` component that's easy to understand and use.

The reason I avoided using `replace-attribute-value` is because there are a bunch of different ways that svgs can be set up and `replace-attribute-value` wouldn't work for all of them. Some use `stroke`, some use `fill`, some have those properties set at the top-level, some have them set on individual child parts of the svg, and some aren't meant to be recolored at all (like the spotify logo). So part of this process involves manually inspecting/updating the created component anyway and I think `replace-attribute-value` wouldn't save us any time. In fact I think it would probably end up creating extra work most of the time

package.json
aahill50: I initially was using the tsx option, but the app was having trouble parsing the components for some reason and I couldn't figure out exactly what the issue was. I think there's probably some issue with the `babel` config that svgr uses, so I'll play around with that and see if I can figure it out. 

I can definitely add title, role and memoize the component

src/assets/CONTRIBUTING.md
aahill50: Ok forget what I said about not using `replace-attribute-value`, I just updated the config to replace any uses of `#000` with the `color` prop. It still doesn't work in every case, but it takes care of some of the work we'd need to do anyway.

package.json
aahill50: Ok, I was able to resolve the `tsx` issue by writing a custom template in the svgr config and fixing the imports/typing etc. I also added `role: 'img'` to the svgProps and `forwardRef` and `memo` to the template

src/assets/formatted/AccountCircle.tsx
```diff
@@ -0,0 +1,47 @@
...
+interface SvgAccountCircleProps {
+  color?: string;
+  height?: number | string;
+  width?: number | string;
+}
+
+const SvgAccountCircle = memo(
+  forwardRef((props: SvgAccountCircleProps, ref: Ref<SVGSVGElement>) => {
+    const { color, height, width } = props;
+    const title = 'SvgAccountCircle';
```
scottdickerson, line 13: I was hoping the consumer could set the `title` as a prop since they'll need to translate it in certain cases.  The titleId might need to be passed as well as an `id` since there might be multiple account circles on the same page.

scottdickerson (commented):
could we leave the Icon story specific one Icon component, and add a new page to the Design section with the Icon grid in it?  `Icon`

image


I think that will make more sense to the Design team

scottdickerson (changes requested):
just a few requested changes

scottdickerson (approved):
LGTM

PR #569, opened 2025-03-29
feat(AddToCalendar): L3-5753 add to calendar component

SRangineni88:
**Jira ticket**

[L3-5753](https://phillipsauctions.atlassian.net/browse/L3-5753)

**Screenshots**
| Before | After |
| ------ | ----- |
image

**Figma link**

[Figma Link](https://www.figma.com/design/OvBXAq48blO1r4qYbeBPjW/RW---Sale-Page-(PLP)?node-id=4791-30655)

**Summary**

Added AddToCalendar component to consume in Remix

**Regression Test**

**Evidence of testing**

- Post logs, screenshots, etc

New Components

[L3-5753]: https://phillipsauctions.atlassian.net/browse/L3-5753?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


scottdickerson:
it looks like it's supposed to flip to the close button when expanded? 
image

scottdickerson:
the padding between the bottom of the button and the popout don't match: 
image

scottdickerson:
the ul/li need to be rendered in a portal so they don't push down other content on the page when expanded.  I think using this Radix widget it probably the best idea as it will also handle accessibility concerns: https://www.radix-ui.com/primitives/docs/components/menubar

scottdickerson (changes requested):
need to handle accessibility with RadixUI

SRangineni88:
> it looks like it's supposed to flip to the close button when expanded? image

Figma screenshot on ticket shows calendar button
image

SRangineni88:
> the ul/li need to be rendered in a portal so they don't push down other content on the page when expanded. I think using this Radix widget it probably the best idea as it will also handle accessibility concerns: https://www.radix-ui.com/primitives/docs/components/menubar

I don't think it will push the content down. Works something like this on calendar page
image

scottdickerson:
I'm worried, not just about accessibility but what if `addtocalendar.com` goes down.  Could we jsut use this library to generate our own ics without leaving our site? https://github.com/nwcell/ics.js/?tab=readme-ov-file

scottdickerson:
a few different javascript libraries that seem widely used
https://snyk.io/advisor/npm-package/ical-generator
https://snyk.io/advisor/npm-package/ics

scottdickerson:
could also use the Radix-ui popover component to achieve accessibility: 
image

scottdickerson:
an example of @zhenpanda using the popover: https://github.com/PhillipsAuctionHouse/seldon/pull/589/files

SRangineni88:
> could also use the Radix-ui popover component to achieve accessibility in combination with those `ics` generator libraries

Sure I will look into this packages

PR #600, opened 2025-04-23
feat(AddToCalendar): L3-5753 Add to calendar component revised

SRangineni88:
**Jira ticket**

[L3-5753](https://phillipsauctions.atlassian.net/browse/L3-5753)

**Screenshots**
| Before | After |
| ------ | ----- |
image | 
image
|

**Figma link**

[Figma Link](https://www.figma.com/design/OvBXAq48blO1r4qYbeBPjW/RW---Sale-Page-(PLP)?node-id=4791-30655)

**Summary**

Create Add to Calendar Component to integrate with Remix Sale page.

Created Calendar button using Radix UI popover and ics library to create downloadable calendar links for outlook and iCalendar.

**Regression Test**

**Evidence of testing**

- Post logs, screenshots, etc

New Components

[L3-5753]: https://phillipsauctions.atlassian.net/browse/L3-5753?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/AddToCalendar/AddToCalendar.tsx
```diff
@@ -0,0 +1,82 @@
...
+ *
+ * [Figma Link](https://www.figma.com/design/OvBXAq48blO1r4qYbeBPjW/RW---Sale-Page-(PLP)?node-id=4791-30655)
+ *
+ * [Storybook Link](https://deploy-preview-600--phillips-seldon.netlify.app/?path=/docs/components-addtocalendar--overview)
+ */
+
+const AddToCalendar = forwardRef<HTMLDivElement, AddToCalendarProps>(
+  ({ className, event, linkElement: Component = Link, ...props }, ref) => {
+    const { className: baseClassName, ...commonProps } = getCommonProps(props, 'AddToCalendar');
+    const buttonAriaLabel = event.title ? `Add ${event.title} to calendar` : 'Add to calendar';
```
scottdickerson, line 33: doesn't this need to be translated

SRangineni88: Hmm I will check with Bella or EVB

src/components/AddToCalendar/_addToCalendar.scss
```diff
@@ -0,0 +1,55 @@
+@use '#scss/allPartials' as *;
+
+.#{$px}-icon-button {
```
scottdickerson, line 3: these styles need to be scoped so they don't override globally

src/components/AddToCalendar/_addToCalendar.scss
```diff
@@ -0,0 +1,55 @@
...
+  background: none;
+  border: none;
+  cursor: pointer;
+
+  &:focus-visible {
+    border-radius: 5px;
+    outline: 1px solid $medium-gray;
+  }
+}
+.#{$px}-link--link {
```
scottdickerson, line 13: same here

src/components/AddToCalendar/_addToCalendar.scss
```diff
@@ -0,0 +1,55 @@
...
+  font-size: 1rem;
+  font-variation-settings: 'wght' 400;
+  line-height: 1.5rem;
+  text-decoration: none;
+}
+
+.atcb-list {
```
scottdickerson, line 20: these seem like old styles no longer needed

src/components/AddToCalendar/CalendarLinks.ts
```diff
@@ -0,0 +1,70 @@
+import type { CalendarEvent } from './types';
+import * as ics from 'ics';
+
+const generateGoogleCalendarLink = (event: CalendarEvent): string => {
```
scottdickerson, line 4: since this file isn't a react component, can we make the filename camelcase?

src/components/AddToCalendar/CalendarLinks.ts
```diff
@@ -0,0 +1,70 @@
...
+    location: event.location,
+    startOutputType: 'local',
+    startInputType: 'local',
+  });
+
+  if (error) {
+    console.error('Error creating iCalendar event:', error);
+  } else {
+    const link = document.createElement('a');
+    link.href = `data:text/calendar;charset=utf8,${value}`;
```
scottdickerson, line 54: this is a nitpick, but I find this a little simpler to read

```
import { createEvent} from 'ics';

const event = {
  ...
}

async function handleDownload() {
  const filename = 'ExampleEvent.ics'
  const file = await new Promise((resolve, reject) => {
    createEvent(event, (error, value) => {
      if (error) {
        reject(error)
      }

      resolve(new File([value], filename, { type: 'text/calendar' }))
    })
  })
  const url = URL.createObjectURL(file);

  // trying to assign the file URL to a window could cause cross-site
  // issues so this is a workaround using HTML5
  const anchor = document.createElement('a');
  anchor.href = url;
  ```

File: src/components/AddToCalendar/AddToCalendar.tsx [hunk 7]
```diff
@@ -0,0 +1,82 @@
...
+      <div {...commonProps} className={classnames(baseClassName, className)} {...props} ref={ref}>
+        <Popover.Root open={open} onOpenChange={setOpen}>
+          <Popover.Trigger asChild>
+            <button aria-label={label} className={`${px}-icon-button ${px}-icon-button--small`} type="button">
+              <Icon icon={open ? 'Close' : 'CalendarAlt'} />
+            </button>
+          </Popover.Trigger>
+          <Popover.Portal>
+            <Popover.Content sideOffset={5} align="start" side="bottom">
+              <ul className="atcb-list" role="dialog">
```
scottdickerson, line 46: these styles need to be prefixed with the `baseClassName` in front

src/components/AddToCalendar/AddToCalendar.tsx
```diff
@@ -0,0 +1,82 @@
...
+}
+
+/**
+ * ## Overview
+ *
+ * Overview of AddToCalendar component
+ *
+ * [Figma Link](https://www.figma.com/design/OvBXAq48blO1r4qYbeBPjW/RW---Sale-Page-(PLP)?node-id=4791-30655)
+ *
+ * [Storybook Link](https://deploy-preview-600--phillips-seldon.netlify.app/?path=/docs/components-addtocalendar--overview)
```
scottdickerson, line 28: this is a weird pattern but I think you'll need to update this to not have `deploy-preview` in there.  You should be able to take a look at the other components to see the right hostname

src/components/AddToCalendar/AddToCalendar.stories.tsx
```diff
@@ -0,0 +1,25 @@
...
+const meta = {
+  title: 'Components/AddToCalendar',
+  component: AddToCalendar,
+} satisfies Meta<typeof AddToCalendar>;
+
+export default meta;
+export const Playground = (props: AddToCalendarProps) => <AddToCalendar {...props} />;
+
+// More on writing stories with args: https://storybook.js.org/docs/react/writing-stories/args
+Playground.args = {
```
scottdickerson, line 14: could you make the story viewport a little larger so we don't have to scroll

src/components/AddToCalendar/AddToCalendar.tsx
```diff
@@ -0,0 +1,82 @@
...
+            <button aria-label={label} className={`${px}-icon-button ${px}-icon-button--small`} type="button">
+              <Icon icon={open ? 'Close' : 'CalendarAlt'} />
+            </button>
+          </Popover.Trigger>
+          <Popover.Portal>
+            <Popover.Content sideOffset={5} align="start" side="bottom">
+              <ul className="atcb-list" role="dialog">
+                <li className="atcb-item">
+                  <button className="atcb-item-link" onClick={() => generateCalendarFile(event)}>
+                    iCalendar
```
scottdickerson, line 49: these need to be translatable and the text size needs to scale responsively.  We should be able to wrap this in a `<Text element="span"` component, looks like the Text variant in the Figma is `body2`

scottdickerson:
the ICS generated file seems to be ignoring the TimeZone. Since TZ was set to `America/New York` it should adjust the raw datetimes correctly

image

src/components/AddToCalendar/CalendarLinks.ts
```diff
@@ -0,0 +1,70 @@
+import type { CalendarEvent } from './types';
+import * as ics from 'ics';
+
+const generateGoogleCalendarLink = (event: CalendarEvent): string => {
+  const start = event.start;
+  const end = event.end;
+  const title = encodeURIComponent(event.title);
+  const description = encodeURIComponent(event.description);
+  const link = `https://www.google.com/calendar/render?action=TEMPLATE&text=${title}&dates=${formatDate(start)}/${formatDate(end)}&details=${description}&location=${event.location}&ctz=${event.timezone}`;
```
scottdickerson, line 9: we need to encode these URL parameters it's chopping off the Auction title in many cases since the example has a `&` in it.  We should probably construct these URLs using `URLSearchParams` and the `URL` object instead of just using templatized strings

scottdickerson:
the Yahoo calendar invite does not appear to be using the TZ correctly.  I'm in Austin so the 3:13 PM should be adjusted to 2:13 PM in my calendar

scottdickerson:
I can't seem to get the Outlook Online link to work for me, not sure if you've had better luck

scottdickerson:
try tabbing with the keyboard onto the button and menu.  `Enter` doesn't open the menu up (only space), and I CAN tab to each of the items inside the menu once it's popped, but the focus state between the button and the links is different and inconsistent

src/components/AddToCalendar/_addToCalendar.scss
SRangineni88: I moved styles to be more specific to calendar component

src/components/AddToCalendar/calendarLinks.ts
```diff
@@ -0,0 +1,128 @@
...
+    return Intl.DateTimeFormat().resolvedOptions().timeZone;
+  } catch {
+    console.warn("Unable to determine user's timezone. Using UTC as a fallback.");
+    return 'UTC';
+  }
+};
+
+const createSearchParams = (params: Record<string, string>) => new URLSearchParams(params);
+
+const generateCalendarLink = (calendarType: 'google' | 'outlook' | 'yahoo', event: CalendarEvent) => {
```
constantinehuzenko, line 23: we can create interface with google, outlook, yahoo and use it both for `calendarType` and CALENDAR_BASE_URL typing

constantinehuzenko:
checked both on desktop and mobile - works fine

constantinehuzenko (approved):
have one comment regarding types but not a blocker

src/components/AddToCalendar/AddToCalendar.tsx
```diff
@@ -0,0 +1,93 @@
...
+                    className={`${baseClassName}-atcb-item-link`}
+                    role="button"
+                    onClick={() => generateCalendarFile(event)}
+                  >
+                    <Text variant={TextVariants.body2}>iCalendar</Text>
+                  </button>
+                </li>
+                <li className={`${baseClassName}-atcb-item`}>
+                  <Component className={`${baseClassName}-atcb-item-link`} href={generateGoogleCalendarLink(event)}>
+                    Google Calendar
```
scottdickerson, line 60: do these need to be translated?

scottdickerson:
I think our keyboard navigability here isn't correct, I can't tab OUT of the menu once I'm in it, it just keeps looping and looping.


https://github.com/user-attachments/assets/41adc6c9-79a5-43f3-9bbe-951c04b7cac0

Technically I think accessibility is supposed to act like this:  https://www.radix-ui.com/primitives/docs/components/dropdown-menu  where the up and down arrows actually switch between the menu items.  Could we replace the dropdown with this?

scottdickerson:
the individual `<li` `buttons` or `links` should have `role=menuitem`

src/components/AddToCalendar/AddToCalendar.tsx
SRangineni88: Spoke to Bella. She said no.

SRangineni88:
> I think our keyboard navigability here isn't correct, I can't tab OUT of the menu once I'm in it, it just keeps looping and looping.
> 
>  Screen.Recording.2025-05-08.at.3.47.26.PM.mov 
> Technically I think accessibility is supposed to act like this: https://www.radix-ui.com/primitives/docs/components/dropdown-menu where the up and down arrows actually switch between the menu items. Could we replace the dropdown with this?

Changing it to dropdown component fixed this loop issue and accessibility. Please take a look.

scottdickerson:
hmm, I still can't tab out of the dropdown menu when it's expanded @SRangineni88 
It should leave the dropdown menu when I tab while it's popped and move to the next focusable element on the page

PR #604, opened 2025-04-28
feat(countdown): L3-6269 added closing state

constantinehuzenko:
**Jira ticket**

[L3-6269](https://phillipsauctions.atlassian.net/browse/L3-6269)

https://github.com/user-attachments/assets/1b46745d-e079-40c5-8b09-d5744af2ef43

**Figma link**

[Figma Link](https://www.figma.com/design/hxqgsE26wM7hII0WaUaDfF/RW---TIMED-Lot-Details--PDP-?t=fbOiFWsATS4NL8uF-0)

**Summary**

Added closing state to the timer
variant={CountdownVariants.compact} was removed because on SalePage in remix it was showing text in black font color and countdown in red which verified by EVB wasn't correct way to display it
the correct way to display everything in red font color

This pull request introduces a new "closing lot" feature to the `Countdown` component, which applies specific styling and behavior when the countdown timer is less than or equal to 3 minutes. It also includes minor cleanup and an unrelated addition to the footer component. Below are the key changes grouped by theme:

### Countdown Component Enhancements:
* Added logic in `Countdown.tsx` to check if the remaining time is exactly or less than 3 minutes, applying the `countdown--closing-lot` class in such cases. (`src/components/Countdown/Countdown.tsx`)
* Introduced new SCSS styles for the `countdown--closing-lot` class to change the color of the countdown label and duration. (`src/components/Countdown/_countdown.scss`)
* Added a new story, `ClosingCountdown`, to showcase the "closing lot" feature in `Countdown.stories.tsx`. (`src/components/Countdown/Countdown.stories.tsx`)

### Testing Improvements:
* Added multiple test cases in `Countdown.test.tsx` to verify the behavior of the `countdown--closing-lot` class under various time conditions (e.g., exactly 3 minutes, less than 3 minutes, more than 3 minutes). (`src/components/Countdown/Countdown.test.tsx`)

### Code Cleanup:
* Removed the unused `CountdownVariants` import and its usage in the `BidSnapshot` component. (`src/patterns/BidSnapshot/BidSnapshot.tsx`) [[1]](diffhunk://#diff-023608454dd8a373f83e678913fe1db898600977d93505f7afa1ff759b3c3e59L9) [[2]](diffhunk://#diff-023608454dd8a373f83e678913fe1db898600977d93505f7afa1ff759b3c3e59L173)

### Miscellaneous:
* Added a "version 1.0.0" text to the footer component. (`src/site-furniture/Footer/Footer.tsx`)

- go to countdown > closing countdown
- verify that when countdown more than 3 minutes it's black color
- verify that when it's 3 or less than 3 minutes it's red color

**Regression Test**

**Evidence of testing**

- Post logs, screenshots, etc

New Components

[L3-6269]: https://phillipsauctions.atlassian.net/browse/L3-6269?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Countdown/Countdown.tsx
+++ b/src/components/Countdown/Countdown.tsx
@@ -93,18 +93,23 @@ const Countdown = forwardRef<HTMLDivElement, CountdownProps>(
       return new Date(endDateTime).getTime() > new Date().getTime();
     }, [endDateTime]);
 
+    const isEqualToThreeMinuses =
+      timeLeft.days === 0 && timeLeft.hours === 0 && timeLeft.minutes === 3 && timeLeft.seconds === 0;
+    const isLessThanThreeMinuses = timeLeft.days === 0 && timeLeft.hours === 0 && timeLeft.minutes < 3;
+
     return showTimer ? (
       <div
         {...commonProps}
         className={classnames(baseClassName, className, {
           [`${baseClassName}--compact`]: variant === CountdownVariants.compact,
           [`${baseClassName}--show-bottom-border`]: showBottomBorder,
+          [`${baseClassName}--closing-lot`]: isEqualToThreeMinuses || isLessThanThreeMinuses,
         })}
         {...props}
         ref={ref}
       >
         <div className={`${baseClassName}__countdown-container`} role="timer" aria-label={label}>
-          <span>{label}</span>
+          <span className={`${baseClassName}__label`}>{label}</span>
           {timeLeft.days > 0 ? (
             <Duration duration={timeLeft} unit="days" locale={dateFnsLocale} formatDurationStr={formatDurationStr} />
           ) : null}
--- a/src/components/Countdown/Countdown.tsx
+++ b/src/components/Countdown/Countdown.tsx
@@ -2,7 +2,13 @@ import { ComponentProps, forwardRef, useEffect, useMemo, useState } from 'react'
 import { getCommonProps } from '../../utils';
 import classnames from 'classnames';
 import { SupportedLanguages } from '../../types/commonTypes';
-import { differenceInDays, differenceInHours, differenceInMinutes, differenceInSeconds } from 'date-fns';
+import {
+  differenceInDays,
+  differenceInHours,
+  differenceInMilliseconds,
+  differenceInMinutes,
+  differenceInSeconds,
+} from 'date-fns';
 import { zhCN, enUS } from 'date-fns/locale';
 import { CountdownVariants } from './types';
 import { Duration } from './Duration'; // Import the Duration component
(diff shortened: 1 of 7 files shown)


src/site-furniture/Footer/Footer.tsx
```diff
@@ -29,8 +29,8 @@ const Footer = ({
 }: FooterProps) => {
   return (
     <footer data-testid={id ? id : `footer`} id={id} className={classnames(`${px}-footer`, { className })}>
+      version 1.0.0
```
adietrich3074, line 32: Just curious. Is this change relevant to the closing state changes

adietrich3074 (approved):
LGTM. Only one comment about the footer version thing

aahill50 (approved):

scottdickerson (changes requested):
small change to our calculation logic requested

src/site-furniture/Footer/Footer.tsx
constantinehuzenko: sorry it was my artifact lol

src/components/Countdown/Countdown.tsx
```diff
@@ -93,18 +99,21 @@ const Countdown = forwardRef<HTMLDivElement, CountdownProps>(
       return new Date(endDateTime).getTime() > new Date().getTime();
     }, [endDateTime]);
 
+    const isClosingTag = differenceInMilliseconds(endDateTime, currentDateTime) <= 3 * 60 * 1000;
```
scottdickerson, line 102: can't we just use `differenceInMinutes` here?

src/components/Countdown/_countdown.scss
```diff
@@ -35,4 +35,11 @@
       gap: $spacing-xsm;
```
scottdickerson, line 35: why we would we want to stop using `--compact` for bidsnapshot timers?  Can you check with @davidicus, it seems like the narrower gap might be important right?

src/components/Countdown/Countdown.tsx
constantinehuzenko: @scottdickerson regarding comparison function
this is our requirement **Timer should change to red when the countdown hits 3 minutes 00 seconds to the close of the lot.**
it means it should turn red when time is `3 minutes 00 second AND BELOW`
when we just use `differenceInMinutes` we will have minutes only like `1`, `2`, `3`
so it will turn countdown red when we have `3 minutes 20 seconds` for example cause it will return just `3`
if we want to turn countdown red from `2 minutes 59 seconds` then we can use `differenceInMinutes`
or we need to add `differenceInSeconds` like I did in prev example

src/components/Countdown/_countdown.scss
scottdickerson: but the `--compact` also narrowed the letter spacing, shouldn't we do that at least? (i.e. shouldn't we fix the `--compact` classes to no longer change the font color instead of removing the compact variant from `BidSnapshot`)

PR #638, opened 2025-06-06
feat(ContentPeek): L3-6647 default ContentPeek to show 100% while loa…

aahill50:
…ding

**Jira ticket**

[L3-6647](https://phillipsauctions.atlassian.net/browse/L3-6647)

**Summary**

The PR adds a fallback value of `none` to the size of the content peek container. This causes the container to fully display until the client side rendering happens, at which point, the `--content-peek-max-height` var gets set and properly renders the ContentPeek. This should address the issue where sometimes the ContentPeek doesn't finish rendering and leaves the content cutoff. Now, if that happens, it should just leave the content fully visible instead of partially cut off. I don't really like the way everything sort of pops in and snaps after loading (see video), so maybe we can look at an improvement for this component in the future to more smoothly handle the load in, but this at least should fix that one issue for now

This pull request includes a minor update to the `_contentPeek.scss` file. The change ensures a fallback value (`100%`) is provided for the `--content-peek-max-height` CSS variable when it is not defined. 

**Regression Test**

**Evidence of testing**

https://github.com/user-attachments/assets/d4c044e5-4499-4324-a76b-9da8e5d5a1ed

New Components

[L3-6647]: https://phillipsauctions.atlassian.net/browse/L3-6647?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/ContentPeek/_contentPeek.scss
+++ b/src/components/ContentPeek/_contentPeek.scss
@@ -10,7 +10,7 @@
     overflow: hidden;
 
     &[data-state='closed'] {
-      max-height: var(--content-peek-max-height);
+      max-height: var(--content-peek-max-height, 100%);
     }
   }
 
--- a/src/components/ContentPeek/_contentPeek.scss
+++ b/src/components/ContentPeek/_contentPeek.scss
@@ -10,7 +10,7 @@
     overflow: hidden;
 
     &[data-state='closed'] {
-      max-height: var(--content-peek-max-height, 100%);
+      max-height: var(--content-peek-max-height, none);
     }
   }
 


zhenpanda (approved):

aahill50:
> I don't think we can show the Read More over top of the image carousel while the ELC is loading

Are you referring to the image in the video? That's an image in the ELC content. The container is still in the same position, so it's not overlapping the carousel

PR #680, opened 2025-08-07
feat(input): L3-6361 Input Adornment

adietrich3074:
**Jira ticket**

[L3-6361](https://phillipsauctions.atlassian.net/browse/L3-6361)

**Screenshots**
Screenshot 2025-08-07 at 1 54 47 PM
Screenshot 2025-08-07 at 1 54 54 PM

**Figma link**

[Figma Link](https://www.figma.com/design/hxqgsE26wM7hII0WaUaDfF/RW---TIMED-Lot-Details--PDP-?m=auto&node-id=12096-400182)

**Summary**

Added the ability to specify an input adornment to preface the input text

This pull request adds support for input adornments to the `Input` component, allowing optional content (such as a symbol or custom node) to be displayed before the input field. The implementation includes changes to the component logic, props, styling, and test coverage.

**Input Adornment Feature**

* Added an optional `inputAdornment` prop to the `Input` component, allowing either a string or a React node to be rendered before the input field. (`src/components/Input/Input.tsx`)
* Updated the `Input` component rendering logic to conditionally display an adornment wrapper and apply appropriate class names when `inputAdornment` is present. (`src/components/Input/Input.tsx`) [[1]](diffhunk://#diff-0414a1d786fc54ea337f45079b8c5d47254051dfc247a8f39d0f4c474fcb9f12R170-R187) [[2]](diffhunk://#diff-0414a1d786fc54ea337f45079b8c5d47254051dfc247a8f39d0f4c474fcb9f12L176-R209)
* Introduced new SCSS styles for the adornment wrapper and its children to ensure proper layout and appearance. (`src/components/Input/_input.scss`)

**Documentation and Storybook**

* Added a new Storybook story `InputWithAdornment` to demonstrate usage of the adornment feature, including example props and layout. (`src/components/Input/Input.stories.tsx`)

**Testing**

* Added a unit test to verify that the input adornment renders correctly when provided, including checks for both the wrapper and custom content. (`src/components/Input/Input.test.tsx`)

- Verify by viewing the [input story](https://phillips-seldon.netlify.app/?path=/docs/components-input--overview) in storybook

**Regression Test**

**Evidence of testing**

- Post logs, screenshots, etc

New Components

[L3-6361]: https://phillipsauctions.atlassian.net/browse/L3-6361?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Input/Input.tsx
+++ b/src/components/Input/Input.tsx
@@ -14,6 +14,11 @@ export interface InputProps extends Omit<React.ComponentProps<'input'>, 'size'> 
   inputAdornment?: string | React.ReactNode;
 
   /**
+   * Optional position to place the adornment. Defaults to 'start'
+   */
+  adornmentPosition?: 'start' | 'end';
+
+  /**
    * Optionally provide the default value of the `<input>`. Should not be passed into controlled input!
    */
   defaultValue?: string | number | readonly string[];
@@ -125,6 +130,7 @@ const Input = React.forwardRef(
     {
       className,
       inputAdornment,
+      adornmentPosition = 'start',
       defaultValue,
       disabled,
       hideLabel,
@@ -168,8 +174,24 @@ const Input = React.forwardRef(
       [`${px}-input--hidden`]: rest.hidden,
     });
 
+    const adornmentInputTypes = [
+      'text',
+      'number',
+      'password',
+      'email',
+      'tel',
+      'url',
+      'search',
+      'date',
+      'datetime-local',
+      'month',
+      'time',
+      'week',
+    ];
+    const isAdornmentInputType = adornmentInputTypes.includes(inputProps.type ?? '');
+
     const inputClassNames =
-      inputAdornment && inputProps.type === 'text'
+      inputAdornment && isAdornmentInputType
         ? classnames(`${px}-input__wrapper__input`, className, { [`${px}-skeleton`]: isSkeletonLoading })
         : classnames(`${px}-input__input`, className, { [`${px}-skeleton`]: isSkeletonLoading });
     const inputPropsToPass = {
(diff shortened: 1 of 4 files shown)


constantinehuzenko:
it breaks down non input components in story, other than that all good
Screenshot 2025-08-11 at 12 36 50 PM

src/components/Input/Input.tsx
```diff
@@ -173,23 +198,16 @@ const Input = React.forwardRef(
...
-          onClick={onClick}
-          placeholder={isSkeletonLoading ? '' : placeholder}
-          readOnly={readOnly}
-          ref={ref}
-          type={inputProps.type}
-          // can't set values on a checkbox or it breaks
-          {...(inputProps.type !== 'checkbox' && inputProps.type !== 'radio' ? { value, defaultValue } : {})}
-          {...rest}
-        />
+        {inputAdornment && inputProps.type === 'text' ? (
```
scottdickerson, line 201: couldn't this be valid for "number" and other types of inputs?  Do we want to limit it?

scottdickerson:
should we allow the adornment to be selected?   Seems like maybe we shouldn't
image

scottdickerson:
looks like in the "error" state (set Invalid to True you'll see what I mean) it no longer highlights the input in red

image

The Figma uses kind of a strange error UX for this field type but I guess that's what they're requesting: 
image

adietrich3074:
> also spotted that we can add couple symbols, I don't think we should allow it Screenshot 2025-08-11 at 4 49 16 PM

Well we don't have any restrictions on what the adornment can be and it should be able to be more than one character. The adornments will be set by the component using the input component so we have control over what is being set. If this was based on user input I could see an issue with this but since we control it, I think we are fine to accept whatever the component sets as the adornment

src/components/Input/Input.tsx
adietrich3074: I do agree that text is very limiting but I don't really see the use case for using adornments with inputs that aren't textbox based. Toggle for one hides it's input tag so you'd never see an adornment anyway. Material UI also only shows examples with a text box. I'll update the conditional to accept more than just the text input type but exclude types like checkbox, radio, toggle etc.

adietrich3074:
> looks like in the "error" state (set Invalid to True you'll see what I mean) it no longer highlights the input in red
> 
> image
> The Figma uses kind of a strange error UX for this field type but I guess that's what they're requesting: image

Waiting to hear back from design on this. The treatment integrated into the design looks to be inline with what we do for all other error states. I want clarification from design before I comment to make a custom error style for this use case

constantinehuzenko (approved):
LGTM

PR #681, opened 2025-08-07
feat(AnnouncementBanner): L3-7452 Create Announcement Bar

zhenpanda:
**Jira ticket**

[Jira Ticket ID](https://phillipsauctions.atlassian.net/browse/L3-7452)

**Screenshots**
Screenshot 2025-08-07 153106

Screenshot 2025-08-15 104350

**Figma link**

[Figma Link]( https://www.figma.com/design/qEd5aXjQeiboeJ8GJMOJba/Global-Notification-Banner?node-id=0-1&p=f&m=dev)

**Summary**

This pull request introduces a new `NotificationBanner` component to the codebase, including its implementation, styling, Storybook story, and tests. It also ensures the component is properly exported and integrated into the project's global styles.

**New NotificationBanner Component:**

* Added a reusable `NotificationBanner` component with props, documentation, and Figma/Storybook references in `NotificationBanner.tsx`.
* Created component-specific styles in `_notificationBanner.scss` for layout and appearance.

**Testing and Documentation:**

* Added Storybook stories for interactive documentation and testing in `NotificationBanner.stories.tsx`.
* Implemented unit tests using common test utilities in `NotificationBanner.test.tsx`.

**Integration and Exports:**

* Exported the component and its types in the component's `index.ts` and the package's main `index.ts` for public use. [[1]](diffhunk://#diff-3281c4b3c4b8d4c814beed7a46eb8ca8512c1ba0e2e97a83447e1d0c1e296157R1) [[2]](diffhunk://#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80R124)
* Included the component's styles in the global `componentStyles.scss` file.

**Regression Test**

**Evidence of testing**

- Post logs, screenshots, etc

New Components


scottdickerson:
can you add to the story this usecase: 
image
Sounds like we need to have it coexist with the SaleAnnouncementBanner in certain cases so we'll need to think about how that would happen

scottdickerson:
> @scottdickerson Announcement is always above the Header while the SaleAnnouncement will always be below the Header https://www.figma.com/design/NtIPFAV2hsMBsuhJrqRW6l/Bidding-Admin-Tool?m=auto&node-id=2832-25222&t=1Cjse11OYpHFy9YY-1

How would we accomplish this on the Remix side? Could we make a story as an example @zhenpanda ?

src/components/NotificationBanner/NotificationBanner.stories.tsx
```diff
@@ -0,0 +1,360 @@
...
+const generateLoremIpsum = (numOfParagraphs = 10) => {
+  let loremIpsum = '';
+  for (let i = 0; i < numOfParagraphs; i++) {
+    loremIpsum += faker.lorem.paragraph();
+  }
+  return loremIpsum;
+};
+
+const fetchData = async (searchQuery: string) => {
+  console.log('searchQuery', searchQuery);
```
DaveMcNamara, line 29: I know it's just a story, but let's remove logs where not useful. Not sure if this case is useful or not!

src/site-furniture/Header/Header.tsx
```diff
@@ -112,7 +124,12 @@ const Header = forwardRef<HTMLElement, HeaderProps>(
     const closeTimeoutRef = React.useRef<NodeJS.Timeout | null>(null);
 
     return (
-      <header {...props} className={classnames(`${px}-header`, className)} ref={ref}>
+      <header
+        {...props}
+        className={classnames(`${px}-header`, className)}
+        ref={ref}
+        style={{ ['top']: `${bannerHeight}px` } as React.CSSProperties}
```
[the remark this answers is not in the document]
davidicus: Just a thought here, but perhaps we just use a fixed container and place all the existing elements inside of the container. They would stay in the flow and we would only have to calculate one item's height when we implement.

@scottdickerson thoughts?

src/site-furniture/Header/Header.tsx
```diff
@@ -111,8 +124,30 @@ const Header = forwardRef<HTMLElement, HeaderProps>(
...
+    const [bannerHeight, setBannerHeight] = useState(bannerRef?.current ? bannerRef.current.clientHeight : 0);
+
+    useEffect(() => {
+      const bannerElement = bannerRef?.current;
+      if (!bannerElement) return;
+
+      setBannerHeight(bannerElement.clientHeight);
+      const resizeObserver = new window.ResizeObserver(() => {
+        setBannerHeight(bannerElement.clientHeight);
+      });
```
DaveMcNamara, line 136: Do we need state? Seems like we could instead set the var directly like

```suggestion
      const resizeObserver = new window.ResizeObserver(() => {
        document.body.style.setProperty("--notification-height", bannerElement.clientHeight)
        // or even just query for the banner's height directly instead of using a ref:
        // document.querySelector(`.${getCommonProps(undefined, 'NotificationBanner').className}`.clientHeight);
      });
```

Then we could remove a number of ref- and style-related props and stuff. This method is probably less React-y but feels much more direct to me. Also I realize `getCommonProps` may be overkill here!

src/site-furniture/Header/Header.tsx
```diff
@@ -111,8 +124,30 @@ const Header = forwardRef<HTMLElement, HeaderProps>(
...
+      };
+    }, [bannerRef, bannerHeight]);
+
     return (
-      <header {...props} className={classnames(`${px}-header`, className)} ref={ref}>
+      <header
+        {...props}
+        className={classnames(`${px}-header`, className)}
+        ref={ref}
+        style={{ '--notification-height': `${bannerHeight}px` } as React.CSSProperties}
```
DaveMcNamara, line 149: I think "notification" and "banner" are used interchangeably, we should probably stick with one name or the other.

src/components/NotificationBanner/NotificationBanner.tsx
```diff
@@ -0,0 +1,34 @@
+import classnames from 'classnames';
+import { ComponentProps, forwardRef } from 'react';
+import { getCommonProps } from '../../utils';
+
+// You'll need to change the ComponentProps<"htmlelementname"> to match the top-level element of your component
```
DaveMcNamara, line 5: Leftover Copilot comment

src/site-furniture/Header/Header.test.tsx
```diff
@@ -115,3 +117,41 @@ describe('useMobileMenu', () => {
...
+
+  function TestHeaderWithBanner({ bannerContentHeight = 0 }: TestHeaderWithBannerProps) {
+    const bannerRef = useRef<HTMLDivElement>(null);
+
+    // Callback ref to mock clientHeight before effect runs
+    const setBannerRef = (el: HTMLDivElement | null) => {
+      if (el) {
+        Object.defineProperty(el, 'clientHeight', {
+          value: bannerContentHeight,
+          configurable: true,
```
DaveMcNamara, line 134: Nitpicky, but why do we need `configurable: true`?

DaveMcNamara (changes requested):
A few thoughts, nothing huge. Please take a look and update the code if any of them make sense to you!

src/components/Search/Search.stories.tsx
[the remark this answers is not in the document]
zhenpanda: Yea this shouldn't be in this PR i'll see to remove it

scottdickerson (approved):
LGTM

DaveMcNamara (approved):
Approved, though I am still a little unsure that the banner height stuff is being handled in an ideal way. Happy to be out-voted, though.

davidicus:
> Approved, though I am still a little unsure that the banner height stuff is being handled in an ideal way. Happy to be out-voted, though.

@DaveMcNamara Agree, it is not. This feature unfortunately needs to be in our next release so we will have to deliver and then change after. We plan on refactoring the header all together

PR #849, opened 2026-04-15
feat(SaleCard): L3-12362 added new child prop

constantinehuzenko:
**Jira ticket**

[L3-12362](https://phillipsauctions.atlassian.net/browse/L3-12362)

**Screenshots**
Screenshot 2026-04-15 at 1 41 41 PM

**Figma link**

[Figma Link](https://www.figma.com/design/RGfh6rH6mWZGr5X0cfn1uz/Global-Registration-2B?node-id=2346-1523&m=dev)

**Summary**

Added new child prop to SaleCard

**Regression Test**

**Evidence of testing**

- Post logs, screenshots, etc

New Components

[L3-12362]: https://phillipsauctions.atlassian.net/browse/L3-12362?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/patterns/SaleCard/SaleCard.stories.tsx
+++ b/src/patterns/SaleCard/SaleCard.stories.tsx
@@ -7,6 +7,7 @@ import { getScssVar } from '../../utils/scssUtils';
 import SaleCard, { SaleCardProps } from './SaleCard';
 import { SaleCardActions } from './SaleCardActions';
 import { Text } from '../../components/Text';
+import { TextVariants } from '../../components/Text/types';
 import { SaleCardImageDisplay, SaleCardVariants } from './types';
 
 const meta = {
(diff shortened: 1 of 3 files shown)


src/patterns/SaleCard/SaleCard.tsx
```diff
@@ -40,6 +40,8 @@ export interface SaleCardProps extends ComponentProps<'article'> {
   variant?: SaleCardVariants;
   /** * The <SaleCardActions /> component used to render the SaleCard CTAs. */
   children?: React.ReactElement<typeof SaleCardActions>;
+  /** * Optional additional content rendered after the info/meta section inside Card.Content. */
+  additionalContent?: React.ReactNode;
```
scottdickerson, line 44: why not just add the additional content react nodes to the children array and find the specific components of type SaleCardActions from the children and put them above whatever else.  Why add an additional prop?

src/patterns/SaleCard/SaleCard.stories.tsx
```diff
@@ -192,6 +193,39 @@ export const SaleCardWithRegistrationStatus = () => (
   </SaleCard>
 );
 
+export const AdditionalContent = () => (
+  <SaleCard
+    {...defaultProps}
+    badgeText="Happening Now"
+    additionalContent={
+      <div
```
scottdickerson, line 201: might be better to have a strong opinion of what this should contain and make a new `SaleCard.AdditionalInfo` component type since our `SaleCard` is an opinionated component

src/patterns/SaleCard/SaleCard.tsx
```diff
@@ -99,6 +102,7 @@ const SaleCard = forwardRef<HTMLElement, SaleCardProps>(
               </div>
             )}
           </Card.Meta>
+          {additionalContent ? additionalContent : null}
```
scottdickerson, line 105: should this also be a new type of `<Card.AdditionalContent>` component or something? just an open question

src/patterns/SaleCard/SaleCard.tsx
smarks8: I looked into this, the children are always rendered as if theyre going to be ctas and hidden on mobile. I think we need to take another look at the card in the future for a refactor if we keep adding on to it but i like the idea of children building blocks

src/patterns/SaleCard/SaleCard.stories.tsx
```diff
@@ -192,6 +193,39 @@ export const SaleCardWithRegistrationStatus = () => (
...
+        style={{
+          alignItems: 'center',
+          borderTop: `1px solid #ECEAE7`,
```
adietrich3074, line 204: Can we use our color token here?

src/patterns/SaleCard/SaleCard.tsx
adietrich3074, line 105: If we don't delineate this as it's own component, or surround this with it's own container that's styled, is there potential for this to break layouts for other variants depending on what's passed in as additional content?

adietrich3074 (changes requested):
Just need some clarification on a couple things

PR #858, opened 2026-04-30
feat(ProgressBar): PDD-256 Build new progress bar component for v2 saleroom

zhenpanda:
**Jira ticket**

[PDD-256](https://phillipsauctions.atlassian.net/browse/PDD-256)

**Screenshots**

https://github.com/user-attachments/assets/7182831e-e45d-467c-8041-a7ee57a5c9c0

**Figma link**

[Figma Link](https://www.figma.com/design/ROSowkNXfQv1nhos5vuyWG/Saleroom-Iteration-V1.1?node-id=3709-102407&m=dev)

**Summary**

Add new progress bar

**Regression Test**

**Evidence of testing**

- Post logs, screenshots, etc

New Components

[PDD-256]: https://phillipsauctions.atlassian.net/browse/PDD-256?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/scss/_vars.scss
```diff
@@ -108,6 +108,8 @@ $bg-soft: #f9f8f7;
 $bg-border: #eae9e6;
 $bg-inverted: #000;
 $bg-overlay: rgba(0, 0, 0, 50%);
+$bg-progress-bar: rgba(0, 101, 252, 20%);
```
smarks8: is this just $status-info at 20%? if so, rather than creating a new global variable can we do something like this in the progressBar.scss https://github.com/PhillipsAuctionHouse/seldon/blob/d5ec326e71773401aac2d235b9bf968b20450c55/src/components/Button/_button.scss#L4

src/scss/_vars.scss
```diff
@@ -108,6 +108,8 @@ $bg-soft: #f9f8f7;
...
+$bg-progress-bar-mobile: #ebf3ff;
```
smarks8, line 113: can you check in with design if this is going to be a new global token?

src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,69 @@
...
+          (1 - 2 * trueProgression) ** 2 * (minVisiblePercent + (100 - minVisiblePercent) * trueProgression) +
+          (1 - (1 - 2 * trueProgression) ** 2) * (100 * trueProgression);
+      }
+
+      return { safeTotal, safeCurrent, visualPercent };
+    }, [currentLot, totalLots]);
+
+    return (
+      <div {...commonProps} ref={ref} className={classnames(baseClassName, className)}>
+        <div
```
smarks8, line 41: i know its a simple component but did you look into using radix or something so we get a lot of accessibility and math ootb? https://www.radix-ui.com/primitives/docs/components/progress

zhenpanda: The issue with using the Progress radix UI is that we still need to do the math ourselves to create the buffer width so that we can see the lot numbers `1/100` etc. If we don't do that the first lot progress width would be a hairline, it's fine if it's `1/10` but some of our auctions have up to like `250` lots 
Below is using the Radix UI Progress bar

Screenshot 2026-04-30 111020

zhenpanda: Screenshot 2026-04-30 111506

src/scss/_vars.scss
zhenpanda: I'll check with design

PR #868, opened 2026-05-12
feat(npmminage): L3-000 npm and node bump, npm package min age

sgrund14:
npm and node bump, min ages for packages

see https://github.com/PhillipsAuctionHouse/phillips-public-remix/pull/2180


package.json
davidicus, line 111:
```suggestion
    "@types/node": "^24.14.1",
```

.npmrc
```diff
@@ -0,0 +1,2 @@
+engine-strict=true
```
scottdickerson, line 1: is there an package exclusion field here we can use or are we waiting to change our package manager?

scottdickerson (changes requested):
might have to remove the min package age until we can get package exclusions working.  TBH I don't foresee `seldon` depending on any other phillips package so we probably could remove that entirely from this PR

PR #882, opened 2026-06-19
feat(TextBanner): PDD-278 update content-wrapper vertical padding

solsaemundsen:
Set `TextBanner` content-wrapper vertical padding to **24px** on mobile/tablet and **32px** on desktop (≥961px), per updated design. Horizontal padding unchanged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/patterns/TextBanner/_textBanner.scss
+++ b/src/patterns/TextBanner/_textBanner.scss
@@ -14,7 +14,7 @@
     flex-direction: column;
     justify-content: space-between;
     max-width: none;
-    padding: $spacing-lg 0;
+    padding: 24px 0;
 
     h1 {
       white-space: nowrap;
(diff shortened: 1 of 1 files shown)


adietrich3074 (approved):

src/patterns/TextBanner/_textBanner.scss
```diff
@@ -33,6 +33,7 @@
 
     @include media($breakpoint-md) {
       flex-direction: row;
+      padding-block: 32px;
```
scottdickerson, line 36: I think we probably should use the legacy spacing tokens here and above, we need to figure out from Christie which tokens will eventually match these sizings.

scottdickerson (changes requested):
legacy tokens request

adietrich3074 (changes requested):
Ignore this comment. I was incorrect about which branch this needed to be PR'd against

PR #895, opened 2026-07-21
feat(component): PDD-683 add slide to activate component

DaveMcNamara:
**Jira ticket**

[PDD-683](https://phillipsauctions.atlassian.net/browse/PDD-683)

**Screenshots**
image

**Figma link**

[Figma Link](https://www.figma.com/design/ROSowkNXfQv1nhos5vuyWG/Saleroom?node-id=8259-82408&m=dev)

**Summary**

Adds `SlideToActivate`, a one-shot slide-to-activate ("swipe to confirm") control for flows like confirming or placing a bid. The thumb can be dragged with a pointer or "charged" and released with the keyboard; on release past the threshold it latches at the end while `onActivation` resolves, then the parent typically marks it complete via `isDisabled` + `disabledReason="complete"`. Supports RTL, two sizes, a configurable corner radius, an optional pill-shaped thumb, and two distinct disabled looks (`blocked` vs `complete`).

- **Add** `SlideToActivate` component (`src/components/SlideToActivate/`):
  - `SlideToActivate.tsx` — root component: track/label/pending indicator, status-driven classnames and `data-status`/`data-disabled-reason` attributes.
  - `components/SlideToActivateThumb.tsx` — the draggable thumb button.
  - `hooks/useSlideToActivate.ts` — state orchestration (progress, status, travel measurement, activation lifecycle).
  - `hooks/useSlideDragHandlers.ts` — pointer-drag gesture (capture, dead zone, sensitivity, threshold-to-activate).
  - `hooks/useSlideKeyboardCharge.ts` — keyboard Enter/Space "hold to charge, release to finish" gesture, with an Escape-to-cancel path.
  - `slideToActivateUtils.ts` — shared constants, easing helpers, travel measurement, and track-click-to-focus helper.
  - `types.ts` — public prop types and enums (`SlideToActivateDirections`, `SlideToActivateBorderRadii`, `SlideToActivateDisabledReasons`; `SlideToActivateSizes` reuses `ButtonSizes` rather than redeclaring it).
  - `index.ts` — public exports.
  - `_slideToActivate.scss` — component styles, including per-radius-token modifiers and reduced-motion handling.
- **Add** Storybook coverage (`SlideToActivate.stories.tsx`, `_slideToActivate.stories.scss`, `SlideToActivateInteractiveDemo.tsx`):
  - 10 stories: `Playground`, `WithThumbIconHeldState`, `PillThumb`, `Widths`, `DisabledStates`, `AsyncActivation`, `Rtl`, `Small`, `CustomStyling`, `ProgressMeter`.
  - Storybook Actions wired for `onActivation`/`onError`/`onProgress`.
  - One interaction (`play`) test on `AsyncActivation`, driving the pointer-drag path in a real browser via `@storybook/addon-vitest` (deliberately not the keyboard path — that ramp is `requestAnimationFrame`-driven and can stall in a backgrounded test tab).
- **Add** unit tests: `SlideToActivate.test.tsx` (49 tests, including full pointer-drag and keyboard-charge/finish animation coverage) and `slideToActivateUtils.test.ts`, plus a new `SlideToActivateInteractiveDemo.test.tsx` (6 tests) for the Storybook demo helper.
- **Fix** a project-wide test-infra gap in `config/vitest/setupTest.ts`: jsdom has no `PointerEvent` constructor, which was silently dropping `pointerId`/`clientX` from any pointer event dispatched via `fireEvent`/`userEvent.pointer` in tests project-wide. Added a minimal `PointerEvent` polyfill (extends `MouseEvent`) to the shared test setup so any component's pointer-interaction tests — not just this one — get real property forwarding. Verified against the full project test suite (see Regression Test).
- **Modify** `src/index.ts` — export the new component.
- **Modify** `src/componentStyles.scss` — import the new component's styles.

1. `npm start` and open **Components/SlideToActivate**.
2. On `Playground`: drag the thumb with a pointer, and separately try holding then releasing Enter/Space — releasing past ~95% travel activates (pending spinner, then latches).
3. Release below the threshold — the thumb snaps back to idle.
4. Check `DisabledStates` (blocked-visible, blocked-hidden, complete), `AsyncActivation` (long pending delay), `Rtl` (mirrored thumb/icon), `Small` (compact size), `PillThumb`/`CustomStyling` (pill shape, branded/compact re-theming), `ProgressMeter` (external bar driven purely by `onProgress`).
5. Confirm keyboard-only usage works end to end (tab to the control, Enter/Space to activate, Escape cancels a hold).

**Regression Test**

- `npm run test` — full unit suite. Since this PR touches the shared `config/vitest/setupTest.ts`, verified the *entire* project suite, not just this component: 116 test files / 1090 tests, all passing, no regressions from the `PointerEvent` polyfill.
- `npm run test-storybook` — runs the new interaction test (requires Playwright browsers installed locally: `npx playwright install`).
- Typecheck/lint/stylelint (`tsc --noEmit`, `eslint`, `stylelint`) — all clean.
- Everything else is purely additive (new export appended in `src/index.ts`, new `@use` appended in `componentStyles.scss`).

**Evidence of testing**

- Typecheck, ESLint, and Stylelint all pass with no errors.
- 55/55 unit tests pass across the three test files.
- Coverage (scoped to this folder) — statements / branches / functions / lines:
  - `SlideToActivate.tsx`, `SlideToActivateThumb.tsx`, `SlideToActivateInteractiveDemo.tsx`, `slideToActivateUtils.ts`: **100 / 100 / 100 / 100**
  - `hooks/useSlideDragHandlers.ts`: 98.66 / 90.9 / 100 / 98.63
  - `hooks/useSlideKeyboardCharge.ts`: 96.15 / 86.53 / 100 / 96
  - `hooks/useSlideToActivate.ts`: 97.56 / 91.66 / 100 / 97.5
  - The handful of remaining uncovered lines are all defensive branches genuinely unreachable through the public component API without contrived setups — e.g. a same-tick reentrancy guard between the keyboard and pointer gesture systems, and a null-ref guard before mount. Documented in code rather than force-tested.
- Full project regression run: 116 test files / 1090 tests passing (relevant since this PR also touches the shared Vitest setup file).
- Manually exercised all 10 stories in a live Storybook session (pointer drag, keyboard charge/release, RTL, disabled variants, async pending state, custom styling, progress-driven external element) and confirmed the interaction test's assertion (`data-status` reaches `pending`) against the real DOM.

- [x] PR title should correctly describe the most significant type of commit. I.e. `feat(scope): ...` if a `minor` release should be triggered. → `feat(components): add SlideToActivate`
- [x] All commit messages follow convention and are appropriate for the changes
- [x] All references to `phillips` class prefix are using the prefix variable (`$px` throughout)
- [x] All major areas have a `data-testid` attribute (via `getCommonProps`)
- [x] Document all props with jsdoc comments (see `types.ts`)
- [x] All strings should be translatable (no hardcoded consumer-facing copy in the component itself — `labelText` etc. are all consumer-supplied)
- [x] Unit tests should be written and should have a coverage of 90% or higher in all areas.

New Components

- [x] Are there any [accessibility considerations](https://www.w3.org/WAI/ARIA/apg/patterns/) that need to be taken into account and tested? — `aria-label`/`aria-busy`/`aria-disabled` on the thumb, a `VisuallyHidden` `role="status"`/`aria-live="polite"` region for pending announcements, keyboard support (Enter/Space/Escape), and `prefers-reduced-motion` handling.
- [x] Default story called "Playground" should be created for all new components
- [x] Create a jsdoc comment that has an Overview section and a link to the Figma design for the component — Overview section is in place; **Figma link still needs to be added** (see placeholder above).
- [x] Export the component and its typescript type from the `index.ts` file
- [x] Import the component scss file into the `componentStyles.scss` file.

[PDD-683]: https://phillipsauctions.atlassian.net/browse/PDD-683?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/SlideToActivate/SlideToActivate.tsx
```diff
@@ -0,0 +1,174 @@
...
+export type { SlideToActivateProps } from './types';
+
+/**
+ * ## Overview
+ *
+ * One-shot slide-to-activate control. Thumb latches at the end while
+ * `onActivation` is pending, then the parent typically marks it complete.
+ *
+ * [Storybook Link](https://phillips-seldon.netlify.app/?path=/docs/components-slidetoactivate--overview)
+ */
```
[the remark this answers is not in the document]
DaveMcNamara: there is no dedicated Figma link for this component

src/components/SlideToActivate/SlideToActivateThumb.tsx
```diff
@@ -0,0 +1,96 @@
+import classnames from 'classnames';
```
scottdickerson, line 1: do we want the extra `components` subdirectory here?

src/components/SlideToActivate/SlideToActivate.stories.demo.tsx
```diff
@@ -0,0 +1,76 @@
+import { useEffect, useState, type ReactNode } from 'react';
```
scottdickerson, line 1: should something like this just be in storybook instead (i.e. a stateful story)

src/components/SlideToActivate/SlideToActivateThumb.tsx
DaveMcNamara: I think it's neater than flat, could call it "subcomponents" to be clearer.

src/components/SlideToActivate/SlideToActivate.stories.demo.tsx
DaveMcNamara: this is only for storybook, this file was borne of the 200-line recommended limit telling me to break stuff out. I thought it was okay here since it isn't exported, but I'll make it more obviously story-only.

src/components/SlideToActivate/SlideToActivateThumb.tsx
DaveMcNamara: I somehow failed to notice that this folder only has one file in it now

scottdickerson:
a few comments from the bot that I think are pretty valid 
```
Prioritized findings                                                                                                                                                                               
                                                                                                                                                                                                     
  ---                                                                                               
  🔴 F1 — The state machine is scattered across three hooks with mirror refs and threaded callbacks. A useReducer would delete a whole category of complexity.                                       
                                                                                                                                                                                                     
  File: hooks/useSlideToActivate.ts + hooks/useSlideDragHandlers.ts + hooks/useSlideKeyboardCharge.ts (~525 lines total)
                                                                                                                                                                                                     
  The current design owns a finite state machine (idle | dragging | pending | snapping) but distributes it across:                                                                                   
                                                                                                                                                                                                     
  - useState for status, progress, maxTravel — for render                                                                                                                                            
  - Mirror refs for the same three values — because event handlers close over stale state (statusRef, progressRef, maxTravelRef)
  - An extra wasDisabledRef for change detection (useSlideToActivate.ts:128-136)                                                                                                                     
  - A snapTimeoutRef for cleanup                                                                                                                                                                     
  - Five callbacks (updateStatus, emitProgress, runActivation, snapToIdle, settleAtEnd) that manually keep state + refs in sync                                                                      
  - Both gesture hooks take those callbacks + refs as options (11-arg config objects) and dispatch by calling them directly                                                                          
  - One special-case cross-hook wire (measureTravel: () => { cancelKeyboardGesture(); measureTravel(); } at useSlideToActivate.ts:158-161) needed because the two peer hooks don't know about each   
  other                                                                                                                                                                                              
                                                                                                                                                                                                     
  This is a distributed mutation pattern — one of the harder-to-reason-about React shapes. The code-judo move here is a useReducer with typed actions:                                               
                                                                                                    
  type State = { status: Status; progress: number; maxTravel: number };                                                                                                                              
  type Action =                                                                                     
    | { type: 'POINTER_START'; originProgress: number }
    | { type: 'POINTER_MOVE'; nextProgress: number }                                                                                                                                                 
    | { type: 'POINTER_END' }                                                                                                                                                                        
    | { type: 'KEY_TICK'; nextProgress: number }                                                                                                                                                     
    | { type: 'KEY_RELEASE' }                                                                                                                                                                        
    | { type: 'ACTIVATE_START' }                                                                                                                                                                     
    | { type: 'ACTIVATE_SETTLED' }              
    | { type: 'ACTIVATE_FAILED' }                                                                                                                                                                    
    | { type: 'SNAP_COMPLETE' }                                                                                                                                                                      
    | { type: 'DISABLED_CHANGED'; isDisabled: boolean }
    | { type: 'MEASURED'; maxTravel: number };                                                                                                                                                       
                                                                                                    
  That collapses:                                                                                                                                                                                    
  - All three mirror refs (state IS the source of truth, dispatch reads it from closure)            
  - The five updateStatus/emitProgress/... callbacks (each event handler dispatches one action)                                                                                                      
  - The wasDisabledRef/isDisabled useEffect (a single DISABLED_CHANGED action atomically resets)
  - The cross-hook cancelKeyboardGesture wrapping around measureTravel (the pointer gesture's POINTER_START reducer clause resets any keyboard-in-progress state in one place)                       
  - Most of the useCallback explosion (6 in the orchestrator, 3+5 in the gesture hooks) — with dispatch, there's little function-identity to preserve                         
                                                                                                                                                                                                     
  Rough sizing: ~200 lines vs the current ~525. The two gesture hooks reduce to event-→-dispatch adapters (~50-60 lines each, no state of their own).                                                
                                                                                                                                                                                                     
  I'd normally flag this as a blocker per the review skill's "prefer solutions that delete complexity" bar. But the current code is correct and has 55 tests locking behavior. So framing this as a  
  strong post-merge follow-up is more honest than blocking — but please open a tracking ticket, don't let this decay into permanent debt. The state model as it stands will be much harder to extend 
  (e.g. adding a "reset to a specific progress" API) than a reducer-based version.                                                                                                                   
                                                                                                    
  ---                                           
  🟠 F2 — 76 lines of copy-pasted &--radius-{token} blocks. A Sass @each collapses this to ~15 lines.
                                                                                                                                                                                                     
  File: _slideToActivate.scss:19-94
                                                                                                                                                                                                     
  Seven blocks, identical modulo the token value:                                                   
                                                                                                                                                                                                     
  &--radius-xs {                                                                                    
    .#{$px}-slide-to-activate__track,           
    .#{$px}-slide-to-activate__thumb-face { border-radius: $radius-xs; }                                                                                                                             
    .#{$px}-slide-to-activate__thumb:focus-visible { @include focus-ring($border-radius: $radius-xs); }                                                                                              
  }                                                                                                                                                                                                  
  &--radius-sm { /* same, with $radius-sm */ }                                                                                                                                                       
  // ... 5 more                                                                                                                                                                                      
                                                                                                    
  This is exactly the "delete a whole category of repetition" opportunity. Replace with:                                                                                                             
  
  $radius-tokens: (xs: $radius-xs, sm: $radius-sm, md: $radius-md, lg: $radius-lg, xl: $radius-xl, '2xl': $radius-2xl, '3xl': $radius-3xl);                                                          
                                                                                                                                                                                                     
  @each $name, $value in $radius-tokens {       
    &--radius-#{$name} {                                                                                                                                                                             
      .#{$px}-slide-to-activate__track,                                                             
      .#{$px}-slide-to-activate__thumb-face { border-radius: $value; }                                                                                                                               
      .#{$px}-slide-to-activate__thumb:focus-visible { @include focus-ring($border-radius: $value); }
    }                                                                                                                                                                                                
  }                                                                                                 
                                                                                                                                                                                                     
  Also — why does a slide control need 7 radius options? Button doesn't expose this many. Suggest matching Button's radius surface (or picking the 2-3 that are actually design-token-approved for   
  this control). Every unused option is a class you have to keep working, snapshot in Chromatic, and reason about.
```

src/components/SlideToActivate/hooks/useSlideDragHandlers.ts
```diff
@@ -0,0 +1,164 @@
...
+      };
+
+      const onPointerUp = (upEvent: PointerEvent) => {
+        if (upEvent.pointerId !== activePointerIdRef.current) {
+          return;
+        }
+        finishDrag();
+      };
+
+      document.addEventListener('pointermove', onPointerMove, { passive: false });
```
scottdickerson: do we want these on the whole document or should we set them on the parent component?

scottdickerson:
d

> a few comments from the bot that I think are pretty valid
> 
> ```
> Prioritized findings                                                                                                                                                                               
>                                                                                                                                                                                                      
>   ---                                                                                               
>   🔴 F1 — The state machine is scattered across three hooks with mirror refs and threaded callbacks. A useReducer would delete a whole category of complexity.                                       
>                                                                                                                                                                                                      
>   File: hooks/useSlideToActivate.ts + hooks/useSlideDragHandlers.ts + hooks/useSlideKeyboardCharge.ts (~525 lines total)
>                                                                                                                                                                                                      
>   The current design owns a finite state machine (idle | dragging | pending | snapping) but distributes it across:                                                                                   
>                                                                                                                                                                                                      
>   - useState for status, progress, maxTravel — for render                                                                                                                                            
>   - Mirror refs for the same three values — because event handlers close over stale state (statusRef, progressRef, maxTravelRef)
>   - An extra wasDisabledRef for change detection (useSlideToActivate.ts:128-136)                                                                                                                     
>   - A snapTimeoutRef for cleanup                                                                                                                                                                     
>   - Five callbacks (updateStatus, emitProgress, runActivation, snapToIdle, settleAtEnd) that manually keep state + refs in sync                                                                      
>   - Both gesture hooks take those callbacks + refs as options (11-arg config objects) and dispatch by calling them directly                                                                          
>   - One special-case cross-hook wire (measureTravel: () => { cancelKeyboardGesture(); measureTravel(); } at useSlideToActivate.ts:158-161) needed because the two peer hooks don't know about each   
>   other                                                                                                                                                                                              
>                                                                                                                                                                                                      
>   This is a distributed mutation pattern — one of the harder-to-reason-about React shapes. The code-judo move here is a useReducer with typed actions:                                               
>                                                                                                     
>   type State = { status: Status; progress: number; maxTravel: number };                                                                                                                              
>   type Action =                                                                                     
>     | { type: 'POINTER_START'; originProgress: number }
>     | { type: 'POINTER_MOVE'; nextProgress: number }                                                                                                                                                 
>     | { type: 'POINTER_END' }                                                                                                                                                                        
>     | { type: 'KEY_TICK'; nextProgress: number }                                                                                                                                                     
>     | { type: 'KEY_RELEASE' }                                                                                                                                                                        
>     | { type: 'ACTIVATE_START' }                                                                                                                                                                     
>     | { type: 'ACTIVATE_SETTLED' }              
>     | { type: 'ACTIVATE_FAILED' }                                                                                                                                                                    
>     | { type: 'SNAP_COMPLETE' }                                                                                                                                                                      
>     | { type: 'DISABLED_CHANGED'; isDisabled: boolean }
>     | { type: 'MEASURED'; maxTravel: number };                                                                                                                                                       
>                                                                                                     
>   That collapses:                                                                                                                                                                                    
>   - All three mirror refs (state IS the source of truth, dispatch reads it from closure)            
>   - The five updateStatus/emitProgress/... callbacks (each event handler dispatches one action)                                                                                                      
>   - The wasDisabledRef/isDisabled useEffect (a single DISABLED_CHANGED action atomically resets)
>   - The cross-hook cancelKeyboardGesture wrapping around measureTravel (the pointer gesture's POINTER_START reducer clause resets any keyboard-in-progress state in one place)                       
>   - Most of the useCallback explosion (6 in the orchestrator, 3+5 in the gesture hooks) — with dispatch, there's little function-identity to preserve                         
>                                                                                                                                                                                                      
>   Rough sizing: ~200 lines vs the current ~525. The two gesture hooks reduce to event-→-dispatch adapters (~50-60 lines each, no state of their own).                                                
>                                                                                                                                                                                                      
>   I'd normally flag this as a blocker per the review skill's "prefer solutions that delete complexity" bar. But the current code is correct and has 55 tests locking behavior. So framing this as a  
>   strong post-merge follow-up is more honest than blocking — but please open a tracking ticket, don't let this decay into permanent debt. The state model as it stands will be much harder to extend 
>   (e.g. adding a "reset to a specific progress" API) than a reducer-based version.                                                                                                                   
>                                                                                                     
>   ---                                           
>   🟠 F2 — 76 lines of copy-pasted &--radius-{token} blocks. A Sass @each collapses this to ~15 lines.
>                                                                                                                                                                                                      
>   File: _slideToActivate.scss:19-94
>                                                                                                                                                                                                      
>   Seven blocks, identical modulo the token value:                                                   
>                                                                                                                                                                                                      
>   &--radius-xs {                                                                                    
>     .#{$px}-slide-to-activate__track,           
>     .#{$px}-slide-to-activate__thumb-face { border-radius: $radius-xs; }                                                                                                                             
>     .#{$px}-slide-to-activate__thumb:focus-visible { @include focus-ring($border-radius: $radius-xs); }                                                                                              
>   }                                                                                                                                                                                                  
>   &--radius-sm { /* same, with $radius-sm */ }                                                                                                                                                       
>   // ... 5 more                                                                                                                                                                                      
>                                                                                                     
>   This is exactly the "delete a whole category of repetition" opportunity. Replace with:                                                                                                             
>   
>   $radius-tokens: (xs: $radius-xs, sm: $radius-sm, md: $radius-md, lg: $radius-lg, xl: $radius-xl, '2xl': $radius-2xl, '3xl': $radius-3xl);                                                          
>                                                                                                                                                                                                      
>   @each $name, $value in $radius-tokens {       
>     &--radius-#{$name} {                                                                                                                                                                             
>       .#{$px}-slide-to-activate__track,                                                             
>       .#{$px}-slide-to-activate__thumb-face { border-radius: $value; }                                                                                                                               
>       .#{$px}-slide-to-activate__thumb:focus-visible { @include focus-ring($border-radius: $value); }
>     }                                                                                                                                                                                                
>   }                                                                                                 
>                                                                                                                                                                                                      
>   Also — why does a slide control need 7 radius options? Button doesn't expose this many. Suggest matching Button's radius surface (or picking the 2-3 that are actually design-token-approved for   
>   this control). Every unused option is a class you have to keep working, snapshot in Chromatic, and reason about.
> ```

more info here 
```
  F1 is that opportunity. It is not implemented in this PR. I'm not blocking on it because the current code works and is tested, but this review is on the record: the current state-machine         
  composition is one obvious refactor away from being dramatically simpler, and letting that decay because "it works now" is the road to permanent debt.
                                                                                                                                                                                                     
  ★ Insight ─────────────────────────────────────                                                                                                                                                    
  - Why useReducer is the right shape here, not XState: the state graph is small (4 nodes, ~11 transitions). A hand-rolled reducer keeps the whole model on one screen. XState earns its keep when
  the graph is either large enough that visualization matters, or when parallel/nested states appear — neither applies here. The reducer + one lifecycle useEffect for runActivation is the sweet    
  spot.                                                                                                                                                                                          
  - Why mirror refs are a code smell in React 18+: the pattern (useState + useRef + manual sync in every setter) usually exists because event handlers registered on document or                     
  requestAnimationFrame close over stale render values. A reducer sidesteps this because dispatch reads state at dispatch time, not at closure time. useSyncExternalStore is another way out for pure
   state-machine cases, but useReducer composes more naturally with React's rendering.                                                                                                               
  - The 19-prop day-one API is a common design-system trap: design systems overweight flexibility on new components ("what if a consumer needs X?"). Actual usage patterns are almost always narrower
   than anticipated, and every option becomes a Chromatic snapshot + jsdoc + test surface you have to keep alive. Ship the intersection of Figma variants and known consumer needs. Add the rest     
  under pressure from a real caller.                                                                                                                                                                 
  ─────────────────────────────────────────────────
```

scottdickerson:
a few more from the bot 
```
The current a11y story is thin on state-change announcements, which is exactly where slide-to-activate controls tend to fail for screen-reader users. Here's what's missing, ranked:
                                                                                                     
  ---                                                                                                                                                                                                
  🔴 The aria-live region only announces the pending state — success and error are silent                                                                                                          
                                                                                                                                                                                                     
  File: SlideToActivate.tsx:126-141                                                                                                                                                                
                                                                                                                                                                                                     
  {isPending ? (                                                                                                                                                                                   
    <>                                                                                                                                                                                               
      <VisuallyHidden asChild>                                                                       
        <span role="status" aria-live="polite">{labelText}</span>                                                                                                                                    
      </VisuallyHidden>                                                                              
      ...                                       
    </>                                                                                                                                                                                            
  ) : null}                                                                                                                                                                                          
                                  
  Two problems:                                                                                                                                                                                      
                                                                                                     
  1. The region unmounts when isPending becomes false. When activation settles (success), the region is gone before it can announce the completed state. A screen reader user who slid to place a bid
   hears "Placing bid…" and then... silence. Never confirms whether the action succeeded.                                                                                                          
  2. On error, activation also transitions pending → idle (via snapToIdle in runActivation's catch block). Same silent unmount. The user has no way to know the action failed and snapped back.      
                                  
  Fix pattern: keep the live region always mounted, just change its content based on the current phase. That way AT reliably picks up transitions:                                                   
                                                                                                                                                                                                     
  <VisuallyHidden asChild>                                                                                                                                                                           
    <span role="status" aria-live="polite" aria-atomic="true">                                                                                                                                       
      {announcementForStatus(status, labelText, /* consumer-supplied messages */)}                                                                                                                   
    </span>                                                                                                                                                                                        
  </VisuallyHidden>                                                                                                                                                                                  
                                                                                                                                                                                                     
  And expose the announcement strings as props so consumers control the wording:                                                                                                                     
  - pendingAnnouncement (default: falls back to labelText)                                                                                                                                           
  - successAnnouncement (required for meaningful UX)                                                 
  - errorAnnouncement (default: something generic like "Action failed, try again")                                                                                                                 
                                                                                                                                                                                                     
  Right now the component's a11y quality depends on the consumer knowing to swap labelText at the exact right times — a leaky contract. The SlideToActivateInteractiveDemo swaps labels correctly, 
  but a naive consumer who just passes labelText="Swipe to confirm" gets no state announcements at all.                                                                                              
                                                                                                                                                                                                   
  ---                                                                                                                                                                                                
  🟠 No documented interaction pattern for keyboard users                                            
                                                                                                                                                                                                     
  The thumb renders as a <button> with aria-label={labelText}. A screen reader tabs to it and hears something like:
                                                
  ▎ "Swipe to confirm, button"                                                                                                                                                                     
                                                                                                                                                                                                     
  That's misleading — it is a button element, but it doesn't behave like one. A normal button activates on press-and-release of Enter/Space. This one requires press-and-hold with a threshold-based 
  release. Nothing in the a11y tree hints at that.                                                                                                                                                   
                                                                                                                                                                                                     
  Fix: add an aria-describedby pointing to a visually hidden hint:                                   
                                                                                                                                                                                                     
  <VisuallyHidden asChild>                                                                           
    <span id={`${id}-instructions`}>            
      Press and hold Enter or Space, then release to activate. Press Escape to cancel.                                                                                                             
    </span>                                                                                                                                                                                          
  </VisuallyHidden>               
                                                                                                                                                                                                     
  <button aria-describedby={`${id}-instructions`} ... />                                                                                                                                             
                                                
  Bonus: aria-keyshortcuts="Enter Space Escape" on the thumb — supported by JAWS/NVDA and hints to the user which keys are relevant.                                                                 
                                                                                                                                                                                                   
  ---                                                                                                                                                                                                
  🟠 Progress during keyboard charge is silent
                                                                                                                                                                                                     
  While the user holds Enter/Space, progress ramps continuously via requestAnimationFrame. AT users get no feedback — they don't know how far along they are, or when they've crossed the threshold.
                                                
  Two ways this could be handled, both defensible:                                                                                                                                                 
                                                                                                                                                                                                     
  1. Skip it — announcing "25%, 50%, 75%…" would be spammy and annoying. Most well-behaved slide-to-activate controls omit this. That's a defensible call if it's a documented decision.             
  2. Announce threshold crossings only — a single "Ready to activate. Release to confirm." when progress crosses requiredProgress gives the user a decision point without spamming.                  
                                                                                                                                                                                                     
  Right now this decision looks incidental (no announcement, no comment explaining why). If it's a deliberate omission, add a code comment; if it's an oversight, threshold-crossing announcements 
  are the right compromise.                                                                                                                                                                          
                                                                                                                                                                                                     
  ---                                                                                                                                                                                                
  🟡 Disabled reason is never communicated                                                                                                                                                           
                                                                                                     
  When isDisabled={true} with disabledReason="blocked" vs "complete", the visual difference is significant (grey/blocked vs black/settled), but AT users get only aria-disabled="true" on the thumb.
  They don't hear why.
                                                                                                                                                                                                   
  Fix: either bake the reason into the aria-label, or use a persistent live-region announcement when disabledReason changes:                                                                         
                                                                                                                                                                                                   
  aria-label={                                                                                                                                                                                       
    isDisabled && disabledReason === 'complete'                                                      
      ? `${labelText}, completed`                                                                                                                                                                    
      : isDisabled && disabledReason === 'blocked'                                                                                                                                                 
        ? `${labelText}, unavailable`                                                                                                                                                                
        : labelText               
  }                                                                                                                                                                                                  
                                                                                                     
  This is especially important because the "complete" state hides the thumb entirely (isThumbHidden) — an AT user tabbing through hits nothing where the interactive element used to be, with no   
  signal that the action was already completed.                                                                                                                                                      
                                                                                                                                                                                                     
  ---                                                                                                                                                                                                
  🟡 Focus management when activation completes                                                                                                                                                      
                                                                                                     
  When the thumb transitions to isDisabled after successful activation, focus is on the disabled thumb. Behavior varies:

  - Some browsers strip focus when an element becomes disabled, leaving focus on <body> (bad — user is lost)                                                                                       
  - Some keep focus on the disabled element (also bad — announced as "disabled, button" with no path forward)                                                                                        
                                                                                                                                                                                                     
  Fix: consumers usually want focus to move to a follow-up action (reset button, next form field). The component can't know where, but it can provide the hook — e.g., accept an onActivated callback
   (fired after onActivation resolves and the component transitions to settled), so consumers get a signal to manage focus. Right now consumers have to observe isDisabled transitions via useEffect,
   which is roundabout.                                                                                                                                                                              
                                                                                                                                                                                                     
  ---                                                                                                                                                                                                
  🟢 Small polish items                                                                              
                                                
  - SlideToActivate.tsx:137 — <Loader aria-label="" /> inside a <span aria-hidden> — the aria-label="" is redundant (the span already hides the loader from AT) and unusual (empty accessible names
  have inconsistent SR behavior). Drop the aria-label="".
  - aria-busy scoping — currently on the thumb only. Consider putting aria-busy={isPending} on the root <div> too, so the whole control is marked busy while an activation is in flight. Small but 
  consistent.                                                                                                                                                                                        
  - prefers-reduced-motion — already handled correctly by useReducedMotion. No gap here.                                                                                                           
  - Track click-to-focus (focusThumbFromTrack) — this is a mouse affordance. AT users don't need it, and it doesn't hurt them. No action needed, but worth noting the pattern is mouse-only.         
                                                                                                                                                                                                   
  ---                                                                                                                                                                                                
  What I'd change in this PR vs. defer                                                                                                                                                               
                                                                                                                                                                                                     
  Must-fix before shipping (real user-visible a11y bugs):                                                                                                                                            
                                                                                                     
  1. Persistent live region + phase-aware content — success and error MUST announce.                                                                                                               

  Fine to defer to a follow-up:

  - Progress announcements during charge (with a comment explaining the deliberate omission)
  - onActivated callback for focus management
  - aria-keyshortcuts

  Just fix while touching:

  - Drop the aria-label="" on the Loader
  - Move the aria-live region to a persistent mount

  ★ Insight ─────────────────────────────────────
  - The core a11y failure mode of slide-to-activate controls is that they look like buttons, get announced like buttons, but don't behave like buttons. Everything else follows from that. The two
  highest-leverage fixes — persistent live region and a description of the interaction pattern — both address that mismatch directly.
  - Live regions that unmount on state change are a very common bug. The intuition ("only show the announcement when it's needed") produces exactly the wrong behavior because AT queues
  announcements from stable live regions. Keep the region always mounted, change its content — that's the pattern.
  - Design-system components should provide a11y mechanisms, not decide a11y content. Success/error announcements are consumer-specific ("Bid placed" vs "Payment confirmed") — expose them as props
  with sensible defaults, don't try to auto-generate. The current design almost does this via labelText swapping, but the coupling is implicit and easy to miss.
  ─────────────────────────────────────────────────

✻ Churned for 1m 36s

❯ do we need to attach these listeners at the document level?                                                                                                                                        

⏺ Short answer: no. The document-level listeners are a mouse-events-era pattern that setPointerCapture obsoleted, and the code is already calling setPointerCapture — so it's using both mechanisms
  redundantly.
                                                                                                                                                                                                     
  What's happening now
```

src/components/SlideToActivate/hooks/useSlideDragHandlers.ts
DaveMcNamara: This is a fallback for some environments where setPointerCapture is messy (JSDom, storybook, old browsers, other weird stuff), looking into moving them to the component level

DaveMcNamara:
> a few comments from the bot that I think are pretty valid
> 
> `...`

I have addressed most of these, but re: radius `@each` collapse: this is required by stylelint's `declaration-property-value-allowed-list`

The `useReducer` refactor is a good call, but it's a big one and will take a little time

zhenpanda (approved):
Nice, working very cool

constantinehuzenko:
We probably need to share preview link with designers and ios and make sure we have the same sensitivity and right feel from this button

src/components/SlideToActivate/slideToActivateUtils.ts
```diff
@@ -0,0 +1,54 @@
+import type { PointerEvent as ReactPointerEvent } from 'react';
+
+export type SlideToActivateStatus = 'idle' | 'dragging' | 'pending' | 'snapping';
+
+export const SNAP_MS = 200;
+/** Hold Enter/Space: ease to full travel; activation still waits for keyup. */
+export const KEYBOARD_CHARGE_TARGET = 1;
+export const KEYBOARD_CHARGE_MS = 900;
+/** Keyup before full charge: burst remaining travel, then activate. */
```
constantinehuzenko: I don't see any point of letting button go to submit because you can't revert it, if you hit button you hit button and that's it right?

constantinehuzenko:
I'll approve once we have green light from design

src/components/SlideToActivate/slideToActivateUtils.ts
DaveMcNamara: Button clicks and keystrokes are usually activated on release, the complication here is that I made it so that holding space/enter with the slider focused will visually slide it (even though releasing space/enter counts as a full slide whether you hold it or not).

zhenpanda (approved):

src/components/SlideToActivate/hooks/useSlideToActivate.ts
```diff
@@ -0,0 +1,198 @@
...
+import { useSlideKeyboardCharge } from './useSlideKeyboardCharge';
+
+export type { SlideToActivateStatus } from '../slideToActivateUtils';
+export { SNAP_MS } from '../slideToActivateUtils';
+
+export interface UseSlideToActivateOptions {
+  requiredProgress: number;
+  deadZone: number;
+  sensitivity: number;
+  direction: 'ltr' | 'rtl';
```
constantinehuzenko: we do have SlideToActivateDirections can we reuse it?

constantinehuzenko (approved):

src/components/SlideToActivate/hooks/useSlideToActivate.ts
DaveMcNamara: yes, that was a miss on my part!

